PR #1, opened 2018-09-07
Write simple mongod tests

Kimchelly:
* Runs multiple mongods concurrently
* Currently only tests blocking processes


process_test.go
```diff
@@ -166,3 +173,171 @@ func TestProcessImplementations(t *testing.T) {
...
+
+func teardownMongods(dbPaths []string) {
+	for _, dbPath := range dbPaths {
+		os.RemoveAll(dbPath)
+	}
+}
+
+func TestMongod(t *testing.T) {
+	if testing.Short() {
+		fmt.Println("SHORT")
```
tychoish: remove this line

process_test.go
```diff
@@ -166,3 +173,171 @@ func TestProcessImplementations(t *testing.T) {
...
+		dbPaths[i] = dbPath
+
+		opts := CreateOptions{Args: []string{mongodPath, "--port", fmt.Sprintf("%d", port), "--dbpath", dbPath}}
+		optslist[i] = opts
+	}
+
+	return optslist, dbPaths, nil
+}
+
+func teardownMongods(dbPaths []string) {
```
tychoish: this function should probably be "removeDBPaths"

process_test.go
```diff
@@ -166,3 +173,171 @@ func TestProcessImplementations(t *testing.T) {
 		})
 	}
 }
+
+// Returns path to release and to mongod
+func downloadMongodb(t *testing.T) (string, string) {
```
tychoish: maybe move these tests and their helpers to a different file. totally optional

process_test.go
Kimchelly: Done

process_test.go
Kimchelly: Done

process_test.go
Kimchelly: Done

PR #2, opened 2018-09-17
Add logging to different endpoints

Kimchelly:
* Can log to different endpoints
* Can configure whether logger reads from stdout and stderr
* Validate() check that output/error redirects aren't cyclical


create.go
```diff
@@ -127,12 +128,60 @@ func (opts *CreateOptions) Resolve(ctx context.Context) (*exec.Cmd, error) {
...
+
+	opts.Output.OutputSender = send.NewWriterSender(outMulti)
+	opts.Output.ErrorSender = send.NewWriterSender(errMulti)
+
 	cmd := exec.CommandContext(ctx, opts.Args[0], args...) // nolint
 	cmd.Dir = opts.WorkingDirectory
-	cmd.Stderr = opts.Output.GetError()
-	cmd.Stdout = opts.Output.GetOutput()
+	cmd.Stdout = io.MultiWriter(opts.Output.GetOutput(), opts.Output.OutputSender)
+	cmd.Stderr = io.MultiWriter(opts.Output.GetError(), opts.Output.ErrorSender)
```
tychoish: shouldn't opts.Output.GetOutput() and opts.Output.OutputSender (ditto for error) be the same thing? or at least, it might be possible to elided Output.OutputSender by moving lines 131-172 into GetOutput/GetError (refactored as needed.)

output.go
```diff
@@ -3,20 +3,58 @@ package jasper
...
-	SendErrorToOutput bool      `json:"redirect_error_to_output"`
+	Output            io.Writer          `json:"-"`
+	Error             io.Writer          `json:"-"`
+	SuppressOutput    bool               `json:"suppress_output"`
+	SuppressError     bool               `json:"suppress_error"`
+	SendOutputToError bool               `json:"redirect_output_to_error"`
+	SendErrorToOutput bool               `json:"redirect_error_to_output"`
+	Loggers           []Logger           `json:"loggers"`
+	OutputSender      *send.WriterSender `json:"-"`
+	ErrorSender       *send.WriterSender `json:"-"`
```
tychoish: is it possible to make these private? or have them be the send.Sender type and then we do the conversion to writer.Sender internally?

output.go
```diff
@@ -3,20 +3,58 @@ package jasper
...
+
+type LogOptions struct {
+	BuildloggerOptions send.BuildloggerConfig    `json:"buildlogger_options"`
+	DefaultPrefix      string                    `json:"default_prefix"`
+	FileName           string                    `json:"file_name"`
+	SplunkOptions      send.SplunkConnectionInfo `json:"splunk_options"`
+	SumoEndpoint       string                    `json:"sumo_endpoint"`
+	// By default, logger reads from both standard output and standard error.
+	IgnoreOutput bool `json:"ignore_log_output"`
+	IgnoreError  bool `json:"ignore_log_error"`
```
tychoish: I think that this should just respect the OutputOptions values.

output.go
```diff
@@ -3,20 +3,58 @@ package jasper
...
+}
+
+type Logger struct {
+	LogType
+	LogOptions
```
tychoish: can we put json tags here *and* name the fields explicitly?

create.go
Kimchelly: I changed these two lines to just call `GetOutput()` and `GetError()` but is it okay that `OutputOptions.Output != OutputOptions.GetOutput()` (and in the same vein, `OutputOptions.Error != OutputOptions.GetError()`)?

output.go
Kimchelly: Done.

output.go
Kimchelly: Done.

jasper.proto
```diff
@@ -5,6 +5,25 @@ option go_package = "internal";
...
+
+enum LogType {
+    LOGBUILDLOGGERV2 = 0;
+    LOGBUILDLOGGERV3 = 1;
+    LOGDEFAULT = 2;
+    LOGFILE = 3;
+    LOGINHERIT = 4;
+    LOGSPLUNK = 5;
+    LOGSUMOLOGIC = 6;
+    LOGUNKNOWN = 7;
```
tychoish: this should probably be 0

jasper.proto
```diff
@@ -5,6 +5,25 @@ option go_package = "internal";
 
 import "google/protobuf/empty.proto";
 
+message Logger {
+    LogType log_type = 1;
+}
+
+message OutputOptions {
+  repeated Logger loggers = 1;
+}
```
tychoish: Won't we need to specify other options?

create.go
tychoish: I think I'm missing a patch

jasper.proto
Kimchelly: I added the other options. Should the LogOptions also be exported?

jasper.proto
Kimchelly: Done.

create.go
Kimchelly: Just pushed it.

jasper_test.go
```diff
@@ -39,27 +39,32 @@ const taskTimeout = 5 * time.Second
 
 func trueCreateOpts() *CreateOptions {
 	return &CreateOptions{
-		Args: []string{"true"},
+		Args:   []string{"true"},
+		Output: OutputOptions{},
```
tychoish: given that this is not a pointer it should be fine let the default constructor take care of it. are you still seeing an error without this line (and friends?)

Kimchelly: I removed OutputOptions{} since it's not needed.

PR #3, opened 2018-09-26
MAKE-472 Add file and in-memory benchmarks

Kimchelly:
* Includes in-memory as logger


benchmark_test.go
```diff
@@ -0,0 +1,124 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	for procName, makeProc := range getProcMap() {
+		b.Run(procName, func(b *testing.B) {
+			for sendTime, timeout := range testDurations() {
+				b.Run(sendTime, func(b *testing.B) {
+					file, err := ioutil.TempFile("build", "bench_out.txt")
+					require.NoError(b, err)
+					defer os.Remove(file.Name())
```
tychoish: I think this means that we're timing the amount of time required to delete the file. 

would it make more sense to create a temporary directory, and then use remove all for the entire directory outside of the scope of the benchmark

tychoish:
a couple of general things: 

- would it be possible to lift, wholesale the godriver's benchmarking harness (this would allow us to output evergreen's legacy perf format,) I imagine that it's probably not quite right, but might be good for checking performance. 
- can we try adding additional larger tests. I think its reasonable to run these for at least 30-90 more seconds in evergreen, so its reasonable to see if we can stress it out.

benchmark_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/send"
+	"github.com/stretchr/testify/require"
+)
+
+type procMaker func(context.Context, *CreateOptions) (Process, error)
+type logMaker func() Logger
+type procMap map[string]procMaker
+type timeMap map[string]time.Duration
```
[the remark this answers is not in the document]
Kimchelly: I removed all of them.

benchmark_test.go
Kimchelly: Nothing is really being timed in the benchmark. The amount of time the process is given depends only on the timeout value, so the testing.B timer isn't used really.

PR #6, opened 2018-10-02
MAKE-457 Download arbitrary files for REST and JRPC

Kimchelly:
* Also fix spelling


jrpc/internal/service.go
```diff
@@ -188,10 +189,28 @@ func (s *jasperService) ResetTags(ctx context.Context, id *JasperProcessID) (*Op
...
+}
+
+func (s *jasperService) DownloadFile(ctx context.Context, info *DownloadInfo) (*OperationOutcome, error) {
+	resp, err := http.Get(info.Url)
+	if err != nil {
+		return &OperationOutcome{Success: false, Text: err.Error()}, errors.Wrap(err, "problem downloading file")
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode >= 300 {
+		err = errors.Errorf("could not download '%s' to path '%s'", info.Url, info.Path)
```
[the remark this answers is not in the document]
Kimchelly: Done.

makefile
```diff
@@ -48,7 +48,8 @@ proto:
 	@mkdir -p jrpc/internal
 	protoc --go_out=plugins=grpc:jrpc/internal *.proto
 clean:
-	rm *.pb.go
+	rm -rf *.pb.go $(buildDir)/bench_out* $(buildDir)/mongodb* $(buildDir)/this_file_exists \
```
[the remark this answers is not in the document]
Kimchelly: I removed these.

rest_service.go
```diff
@@ -337,3 +338,44 @@ func (s *Service) closeManager(rw http.ResponseWriter, r *http.Request) {
 
 	gimlet.WriteJSON(rw, struct{}{})
 }
+
+func (s *Service) downloadFile(rw http.ResponseWriter, r *http.Request) {
+	var downloadInfo struct {
+		URL  string
```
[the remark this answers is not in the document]
Kimchelly: Added them.

jrpc/client.go
```diff
@@ -99,13 +99,24 @@ func (m *jrpcManager) Close(ctx context.Context) error {
...
+func (m *jrpcManager) DownloadFile(ctx context.Context, url string, path string) error {
+	resp, err := m.client.DownloadFile(ctx, &internal.DownloadInfo{Url: url, Path: path})
+	if err != nil {
+		return errors.WithStack(err)
+	}
+	if resp.Success {
+		return nil
+	}
+	return errors.New(resp.Text)
+}
```
tychoish: I think we don't need this here. the jrpcManager type is exported only as an implementation of jasper.Manager, which means this isn't exported.

This operation will be available to clients who are using the generated grpc client (stub), so it's fine to just remove this, I think.

jrpc/internal/service.go
```diff
@@ -188,10 +189,28 @@ func (s *jasperService) ResetTags(ctx context.Context, id *JasperProcessID) (*Op
...
+			Text:    err.Error(),
 		}, err
 	}
 	proc.ResetTags()
-	return &OperationOutcome{Succuess: true, Text: "set tags"}, nil
+	return &OperationOutcome{Success: true, Text: "set tags"}, nil
+}
+
+func (s *jasperService) DownloadFile(ctx context.Context, info *DownloadInfo) (*OperationOutcome, error) {
+	resp, err := http.Get(info.Url)
```
tychoish: should probably let users pass a client to the service in its constructor rather than using the default http package client one.

jrpc/internal/service.go
```diff
@@ -188,10 +189,28 @@ func (s *jasperService) ResetTags(ctx context.Context, id *JasperProcessID) (*Op
...
+	if err != nil {
+		return &OperationOutcome{Success: false, Text: err.Error()}, errors.Wrap(err, "problem downloading file")
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode >= 300 {
```
tychoish: I'd just do != 200

util.go
```diff
@@ -9,3 +15,21 @@ func sliceContains(group []string, name string) bool {
 
 	return false
 }
+
+// WriteFile writes the buffer to the file.
+func WriteFile(reader io.Reader, path string) error {
+	dirPath := filepath.Dir(path)
+	if err := os.MkdirAll(dirPath, os.ModeDir|os.ModePerm); err != nil {
```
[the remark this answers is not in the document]
tychoish: I think I asked for this. I think making enclosing directories is going to be useful in this situation, and it would be annoying to have to do it in any of the use cases where we'd be using this functionality.

util.go
```diff
@@ -9,3 +15,21 @@ func sliceContains(group []string, name string) bool {
...
+		return err
+	}
+
+	file, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY, 0644)
```
tychoish: I think there's an os.Create that does roughly this.

jrpc/client.go
Kimchelly: Removed it.

jrpc/internal/service.go
Kimchelly: Added a client to the jasperService struct.

util.go
Kimchelly: Changed to os.Create

PR #7, opened 2018-10-05
MAKE-456 Download MongoDB using bond and use LRU cache

Kimchelly:
* MongoDB download files automatically added to cache
* Downloads are asynchronous
* Cache can be configured
* For REST and JRPC


jasper_test.go
```diff
@@ -32,7 +32,9 @@ func getPortNumber() int {
 
 }
 
-const taskTimeout = 5 * time.Second
+// const taskTimeout = 5 * time.Second
+// kim: remove when DownloadMongoDB test works
```
bsamek: Does it work?

jrpc/internal/service.go
```diff
@@ -21,16 +24,43 @@ func AttachService(manager jasper.Manager, s *grpc.Server) error {
 	srv := &jasperService{
 		hostID:  hn,
 		manager: manager,
+		cache:   lru.NewCache(),
+		cacheOpts: jasper.CacheOptions{
+			Disabled:   false,
```
bsamek: nit: redundant, since this is the zero value

rest_client.go
```diff
@@ -237,6 +237,54 @@ func (c *restClient) getLogs(id string) ([]string, error) {
...
+		return err
+	}
+
+	req, err := http.NewRequest(http.MethodPost, c.getURL("/download-mongodb"), body)
+	if err != nil {
+		return errors.Wrap(err, "problem building request")
+	}
+	req = req.WithContext(ctx)
+
+	resp, err := c.client.Do(req)
```
bsamek: Does the client handle closing the response body? If not, you may have a resource leak here and in the next function.

rest_service.go
```diff
@@ -35,10 +48,24 @@ func NewManagerService(m Manager) *Service {
...
+	// DefaultMaxCacheSize is the maximum allowed size of the LRU cache.
+	DefaultMaxCacheSize = 1024 * 1024 * 1024
+)
+
 // App constructs and returns a gimlet application for this
 // service. It attaches no middleware and does not start the service.
 func (s *Service) App() *gimlet.APIApp {
 	s.hostID, _ = os.Hostname()
+	s.cache = lru.NewCache()
+	s.cacheMutex.Lock()
```
bsamek: If you don't need to unlock a mutex before the function returns, the Go idiom is:
```
mu.Lock()
defer mu.Unlock()
```
It makes it a bit easier to not forget to unlock the mutex.

rest_service.go
```diff
@@ -54,10 +81,32 @@ func (s *Service) App() *gimlet.APIApp {
...
 }
 
+func (s *Service) backgroundPrune() {
+	s.cacheMutex.RLock()
+	timer := time.NewTimer(s.cacheOpts.PruneDelay)
+	s.cacheMutex.RUnlock()
+
+	for {
+		<-timer.C
+		s.cacheMutex.RLock()
```
bsamek: Do we expect that `cacheOpts` will change after they're set?

rest_service.go
```diff
@@ -394,3 +443,239 @@ func (s *Service) closeManager(rw http.ResponseWriter, r *http.Request) {
...
+		catcher := grip.NewBasicCatcher()
+
+		filePath := filepath.Join(absRootPath, fileName)
+		if err := cache.AddFile(filePath); err != nil {
+			catcher.Add(errors.Wrap(err, "problem adding file to cache"))
+		}
+
+		baseName := filepath.Base(fileName)
+		ext := filepath.Ext(baseName)
+		dirPath := filepath.Join(absRootPath, string(baseName[:len(baseName)-len(ext)]))
```
bsamek: Nit: Could be cleaner with `strings.TrimSuffix`.

rest_service.go
```diff
@@ -394,3 +443,239 @@ func (s *Service) closeManager(rw http.ResponseWriter, r *http.Request) {
...
+			// Cache only handles individual files, not directories.
+			if !info.IsDir() {
+				if err := cache.AddStat(path, info); err != nil {
+					catcher.Add(errors.Wrap(err, "problem adding file to cache"))
+				}
+			}
+
+			return nil
+		})
+		if err != nil {
```
bsamek: This check is unnecessary. `errors.Wrap(nil)` returns nil and `catcher.Add(nil)` is a noop. You can just `catcher.Add(errors.Wrap(err, "problem walking download files"))` without the `if`.

rest_service.go
```diff
@@ -394,3 +443,239 @@ func (s *Service) closeManager(rw http.ResponseWriter, r *http.Request) {
...
+	}
+	if err := opts.Validate(); err != nil {
+		writeError(rw, gimlet.ErrorResponse{
+			StatusCode: http.StatusBadRequest,
+			Message:    errors.Wrap(err, "problem with validating cache options").Error(),
+		})
+		return
+	}
+
+	s.cacheMutex.Lock()
```
bsamek:
```
s.cacheMutex.Lock()
defer s.cacheMutex.Unlock()
```

rest_client.go
Kimchelly: I added the close.

rest_service.go
Kimchelly: Changed to defer.

rest_service.go
Kimchelly: Yes. At the moment, the client can only change cacheOpts by using the `/configure-cache` route, so `backgroundPrune()` is already running by the time the client has a chance to set their own options.

rest_service.go
Kimchelly: Done.

PR #8, opened 2018-10-05
MAKE-476 Buffered sender configuration

Kimchelly:


output.go
```diff
@@ -120,59 +129,61 @@ func (l LogType) Configure(opts LogOptions) (*send.Sender, error) {
 		if opts.BuildloggerOptions.Local.Name() == "" {
 			return nil, errors.New("must call SetName() on buildlogger local sender")
 		}
-		s, err := send.MakeBuildlogger(DefaultLogName, &opts.BuildloggerOptions)
+		sender, err = send.MakeBuildlogger(DefaultLogName, &opts.BuildloggerOptions)
```
Kimchelly: This assignment (and all the similar changes beneath this) doesn't change functionality at all but I think it looks cleaner than the way I originally wrote it.

PR #9, opened 2018-10-08
MAKE-485 Download arbitrary files asynchronously

Kimchelly:
* Also refactor some common logic between synchronous and asynchronous downloading into helper functions


async_test.go
```diff
@@ -170,3 +171,33 @@ func TestAddMongoDBFilesToCacheWithBadPath(t *testing.T) {
...
+	defer os.Remove(file.Name())
+
+	info := DownloadInfo{
+		URL:  "https://example.com",
+		Path: file.Name(),
+	}
+	req, err := http.NewRequest(http.MethodGet, info.URL, nil)
+	require.NoError(t, err)
+
+	assert.NoError(t, DoDownload(req, info, http.Client{}))
```
bsamek: If this should write a thing to the filesystem, should you check that it really did?

rest_client.go
```diff
@@ -55,6 +55,25 @@ func handleError(resp *http.Response) error {
...
+		return nil, errors.Wrap(err, "problem building request")
+	}
+
+	req = req.WithContext(ctx)
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return nil, errors.Wrap(err, "problem making request")
+	}
+	if err = handleError(resp); err != nil {
+		defer resp.Body.Close()
```
bsamek: Why only close in the error case?

Kimchelly: I assumed that the successful server response status was 200 ok. Returning an error on line 71 means the server returned non-200 so the caller of `doRequest` is not responsible for closing the response body since the server did not successfully perform the operation. Returning nil error means the server returned 200 and the caller can use the response object (so they're also responsible for closing the response).

At the moment, the user is only responsible for closing `resp.Body` if there was no error. If it's cleaner, I could just move the handleError portion out of this function and then the user would always be responsible for closing the response body.

async_test.go
Kimchelly: Added a check for non-empty file.

rest_client.go
bsamek: Ah, gotcha, that makes sense. LGTM.

PR #14, opened 2018-10-11
MAKE-489 Add log formatter option

Kimchelly:


output.go
```diff
@@ -44,12 +44,22 @@ const (
 	DefaultLogName = "jasper"
 )
 
+type LogFormat string
+
+const (
+	LogFormatPlain   = ""
```
tychoish: we should make this something more than the empty string

output.go
```diff
@@ -159,14 +178,29 @@ func (l LogType) Configure(opts LogOptions) (send.Sender, error) {
...
+	switch opts.Format {
+	case LogFormatDefault:
+		formatter = send.MakeDefaultFormatter()
+	case LogFormatPlain:
+		formatter = send.MakePlainFormatter()
+	case LogFormatJSON:
+		formatter = send.MakeJSONFormatter()
+	default:
+		return nil, errors.New("unknown log format")
+	}
```
tychoish: should we pull this out into a method on the formatter type?

jasper.proto
```diff
@@ -30,14 +30,22 @@ enum LogType {
...
     BuildloggerOptions buildlogger_options = 2;
     string default_prefix = 3;
     string file_name = 4;
-    int64 in_memory_cap = 5;
-    SplunkOptions splunk_options = 6;
-    string sumo_endpoint = 7;
+    string format = 5;
+    int64 in_memory_cap = 6;
+    SplunkOptions splunk_options = 7;
+    string sumo_endpoint = 8;
```
tychoish: fyi: this is fine because we know nothing is depending on it, but in general we can't change field numbers without breaking compatibility.

output.go
Kimchelly: Changed to "plain".

output.go
Kimchelly: I moved this to its own LogFormat method.

jasper.proto
Kimchelly: 👍

PR #15, opened 2018-10-12
MAKE-387 Clean up download tests and require absolute paths for downloads

Kimchelly:
* Also move download-related functions from rest_service.go to download.go
* More tests for download

### DIFF
--- a/download.go
+++ b/download.go
@@ -1,0 +1,314 @@
+package jasper
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"time"
+
+	"github.com/mholt/archiver"
+	"github.com/mongodb/amboy"
+	"github.com/mongodb/amboy/queue"
+	"github.com/mongodb/grip"
+	"github.com/pkg/errors"
+	"github.com/tychoish/bond"
+	"github.com/tychoish/bond/recall"
+	"github.com/tychoish/lru"
+)
+
+// DownloadInfo represents the URL to download and the file path where it should be downloaded.
+type DownloadInfo struct {
+	URL         string         `json:"url"`
+	Path        string         `json:"path"`
+	ArchiveOpts ArchiveOptions `json:"archive_opts"`
+}
+
+// Validate checks the download options.
+func (info DownloadInfo) Validate() error {
+	catcher := grip.NewBasicCatcher()
+
+	if !filepath.IsAbs(info.Path) {
+		catcher.Add(errors.New("download path must be an absolute path"))
+	}
+
+	catcher.Add(info.ArchiveOpts.Validate())
+
+	return catcher.Resolve()
+}
+
+// ArchiveFormat represents an archive file type.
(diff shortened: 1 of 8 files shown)


download.go
```diff
@@ -0,0 +1,314 @@
...
+		if catcher.HasErrors() {
+			errOut <- catcher.Resolve()
+		}
+		close(errOut)
+	}()
+
+	return output, errOut
+}
+
+func aggregateErrors(groups ...<-chan error) error {
```
tychoish: you can just pass catcher around, they're threadsafe.

download.go
```diff
@@ -0,0 +1,314 @@
...
+	return nil
+}
+
+func addMongoDBFilesToCache(cache *lru.Cache, absRootPath string) func(string) error {
+	return func(fileName string) error {
+		catcher := grip.NewBasicCatcher()
+
+		filePath := filepath.Join(absRootPath, fileName)
+		if err := cache.AddFile(filePath); err != nil {
+			catcher.Add(errors.Wrapf(err, "problem adding file %s to cache", filePath))
```
tychoish: should return rather than catch things.

download.go
```diff
@@ -0,0 +1,314 @@
...
+		return errors.Wrap(err, "problem starting download job queue")
+	}
+
+	if err := amboy.PopulateQueue(ctx, q, jobs); err != nil {
+		return errors.Wrap(err, "problem adding download jobs to queue")
+	}
+
+	go func() {
+		if err := processJobs(q); err != nil {
+			grip.Errorf(errors.Wrap(err, "error occurred while adding jobs to cache").Error())
```
tychoish: what do we do this this error?

jrpc/internal/service.go
```diff
@@ -232,12 +232,14 @@ func (s *jasperService) ResetTags(ctx context.Context, id *JasperProcessID) (*Op
 
 func (s *jasperService) DownloadMongoDB(ctx context.Context, opts *MongoDBDownloadOptions) (*OperationOutcome, error) {
 	jopts := opts.Export()
-	if err := jopts.BuildOpts.Validate(); err != nil {
-		return &OperationOutcome{Success: false, Text: errors.Wrap(err, "invalid build options").Error()}, err
+	if err := jopts.Validate(); err != nil {
+		err = errors.Wrap(err, "problem validating MongoDB download options")
+		return &OperationOutcome{Success: false, Text: err.Error()}, err
```
tychoish: I would just combine this here and below into one longer line,

download.go
Kimchelly: Done.

download.go
Kimchelly: Changed to use catcher.

jrpc/internal/service.go
Kimchelly: Done. This is done in several other places. Do you want them all changed like this?

download.go
Kimchelly: Truthfully, not much. It's only there to print a message in case an error occurs while it's asynchronously downloading.

PR #18, opened 2018-12-06
MAKE-556: Remove basicProcess from Jasper, some other fixes

Utagai:
`basicProcess` was broken and there wasn't much of a way to fix it. It also was not ever really used, so it has been removed from Jasper. Instead, `blockingProcess` can and should be used, sometimes wrapped with a `localProcess`.

Removing `basicProcess` then revealed some issues with `blockingProcess` and the Jasper REST service, which have also been fixed.

~This is not actually related to any ticket on JIRA~ ([MAKE-565](https://jira.mongodb.org/browse/MAKE-565)), but were issues found when investigating [MAKE-520](https://jira.mongodb.org/browse/MAKE-520).

Some of these issues are not related to MAKE-565, but were small enough to include in this PR anyways.

There is still a race with REST that needs to be fixed, and occurs somewhat rarely but definitely exists, but this clearly has existed on master for quite some time now, so it is not part of this PR for removing `basicProcess`.


harness.go
```diff
@@ -30,13 +30,6 @@ func yesCreateOpts(timeout time.Duration) CreateOptions {
 	return CreateOptions{Args: []string{"yes"}, Timeout: timeout}
 }
 
-func procMap() map[string]func(context.Context, *CreateOptions) (Process, error) {
-	return map[string]func(context.Context, *CreateOptions) (Process, error){
-		"Blocking": newBlockingProcess,
-		"Basic":    newBasicProcess,
-	}
```
tychoish: I'd leave this as it is so that we can add a different implementation in the future.

Utagai: Done.

PR #20, opened 2018-12-17
MAKE-528: add methods to clean up/clear terminated processes, also remove some redundant code

Utagai:
I've opted to go for a design choice where we expose a `Clear()` to users if they want to use it, and also provide a `selfClearingProcessManager` that clears itself (and lazily, so that we don't have to pay extra cost clearing every time).


manager_self_clearing_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func (m *selfClearingProcessManager) registerBasedCreate(ctx context.Context, t *testing.T, opts *CreateOptions) (Process, error) {
```
tychoish: it's a little weird to define methods on a type in the test file, at least to my eyes.

manager_local.go
```diff
@@ -15,14 +15,6 @@ func NewLocalManager() Manager {
 	}
 }
 
-func NewLocalManagerBlockingProcesses() Manager {
```
tychoish: why delete?

interface.go
```diff
@@ -26,6 +26,7 @@ type Manager interface {
 	List(context.Context, Filter) ([]Process, error)
 	Group(context.Context, string) ([]Process, error)
 	Get(context.Context, string) (Process, error)
+	Clear(context.Context) error
```
tychoish: are there actually cases where this can return an error?

manager_self_clearing_test.go
Utagai: I agree. I also felt weird about this, but it helped remove duplication in the test code.

manager_local.go
Utagai: This was only here because we used to have `basicProcess`, and so there was a need to distinguish a normal `NewLocalManager` and a `NewLocalManagerBlockingProcesses`. Since we removed `basicProcess`, this distinction is no longer necessary.

interface.go
Utagai: Not many, however, this exists because sometimes it can, but mostly in the REST implementation, since we have code there to handle errors with requests and such (https://github.com/mongodb/jasper/pull/20/files#diff-b215740dd688afb2721c9daf2c1287e3R216).

PR #21, opened 2018-12-19
MAKE-520: Retrieve the exit code from calling Wait() in jasper

Utagai:
Added a self-review comment to this patch because there is something in particular that bothers me about what we are doing here.

### DIFF
--- a/jrpc/client.go
+++ b/jrpc/client.go
@@ -174,17 +174,20 @@ func (p *jrpcProcess) Signal(ctx context.Context, sig syscall.Signal) error {
 	return errors.New(resp.Text)
 }
 
-func (p *jrpcProcess) Wait(ctx context.Context) error {
+func (p *jrpcProcess) Wait(ctx context.Context) (int, error) {
 	resp, err := p.client.Wait(ctx, &internal.JasperProcessID{Value: p.info.Id})
 	if err != nil {
-		return errors.WithStack(err)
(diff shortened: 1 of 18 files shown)


jrpc/client.go
```diff
@@ -170,17 +170,20 @@ func (p *jrpcProcess) Signal(ctx context.Context, sig syscall.Signal) error {
...
+func (p *jrpcProcess) Wait(ctx context.Context) (int, error) {
 	resp, err := p.client.Wait(ctx, &internal.JasperProcessID{Value: p.info.Id})
 	if err != nil {
-		return errors.WithStack(err)
+		return -1, errors.WithStack(err)
 	}
 
 	if resp.Success {
-		return nil
+		if resp.ExitCode != 0 {
```
Utagai: This check against the `ExitCode`/`exitCode` value is something we do also in [`rest_service.go`](https://github.com/mongodb/jasper/pull/21/files#diff-67a967463fc3e0e882fb1fcb3f1c2156R175).

This is something I don't think is particular robust across all platforms, since we don't have a guarantee that 0 is successful and otherwise it is not. I think if we want to fix this, we need to return extra information from `Wait()`. This either means we return more "full" information, like the entire `ProcessState`, or we change the signature to also return a `bool` indicating whether the process terminated successfully or not. This isn't impossible, but I believe this is the only thing we can do if we want to avoid making a call to `p.Info()`.

PR #23, opened 2018-12-19
MAKE-571: Increase timeout for Jasper benchmarks (also check that they pass)

Utagai:
This also adds some logic that ensures that the benchmark actually passes.

We may want to prefer using "^FAIL" instead here, just to be consistent with MAKE-587.

### DIFF
--- a/makefile
+++ b/makefile
@@ -30,7 +30,8 @@ test:
 .PHONY: benchmark
 benchmark:
 	@mkdir -p $(buildDir)
-	go test $(testArgs) -bench=$(benchPattern) $(if $(RUN_TEST),, -run=^^$$) | tee $(buildDir)/bench.sink.out
+	go test $(testArgs) -timeout 45m -bench=$(benchPattern) $(if $(RUN_TEST),, -run=^^$$) | tee $(buildDir)/bench.sink.out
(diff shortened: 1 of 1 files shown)


makefile
```diff
@@ -30,7 +30,8 @@ test:
 .PHONY: benchmark
 benchmark:
 	@mkdir -p $(buildDir)
-	go test $(testArgs) -bench=$(benchPattern) $(if $(RUN_TEST),, -run=^^$$) | tee $(buildDir)/bench.sink.out
+	go test $(testArgs) -timeout 45m -bench=$(benchPattern) $(if $(RUN_TEST),, -run=^^$$) | tee $(buildDir)/bench.sink.out
```
tychoish: I'd put this on line 7 rather than here.

Utagai: I avoided this because `testArgs` is used in more than just the `benchmark` target, and is used in the other targets. This would cause even the non-benchmark tasks to timeout after 45 minutes instead of the default of 10. This doesn't seem desirable since the normal tests should definitely finish in under 10 minutes.

We could always add a `benchArgs` or something, specifically for benchmarks, but it would only ever be used once in the `benchmark` target.

PR #24, opened 2018-12-21
MAKE-525: Errors with sending signal to terminated processes, some other clean up

Utagai:
I've decided to just change the error returns for `Wait()` and `Signal()` to `nil` so that their `OperationOutcome` returns actually make it through gRPC.

It may be worth it to do something similar to other times where we return `(*OperationOutcome, error)` and its better to send over the `*OperationOutcome` instead.

### DIFF
--- a/jrpc/internal/service.go
+++ b/jrpc/internal/service.go
@@ -141,7 +141,7 @@ func (s *jasperService) Signal(ctx context.Context, sig *SignalProcess) (*Operat
 			Success:  false,
 			Text:     err.Error(),
 			ExitCode: -2,
-		}, err
+		}, nil
 	}
 
 	if err = proc.Signal(ctx, sig.Signal.Export()); err != nil {
@@ -150,7 +150,7 @@ func (s *jasperService) Signal(ctx context.Context, sig *SignalProcess) (*Operat
 			Success:  false,
 			ExitCode: -3,
 			Text:     err.Error(),
-		}, err
+		}, nil
 	}
 
 	return &OperationOutcome{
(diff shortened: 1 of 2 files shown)


jrpc/internal/service.go
```diff
@@ -286,7 +286,10 @@ func (s *jasperService) ResetTags(ctx context.Context, id *JasperProcessID) (*Op
 func (s *jasperService) DownloadMongoDB(ctx context.Context, opts *MongoDBDownloadOptions) (*OperationOutcome, error) {
 	jopts := opts.Export()
 	if err := jopts.Validate(); err != nil {
-		return &OperationOutcome{Success: false, Text: errors.Wrap(err, "problem validating MongoDB download options").Error()}, errors.Wrap(err, "problem validating MongoDB download options")
+		return &OperationOutcome{
+			Success: false,
+			Text:    errors.Wrap(err, "problem validating MongoDB download options").Error(),
+		}, errors.Wrap(err, "problem validating MongoDB download options")
```
tychoish: should this be nil error?

jrpc/internal/service.go
```diff
@@ -301,7 +304,10 @@ func (s *jasperService) ConfigureCache(ctx context.Context, opts *CacheOptions)
 	jopts := opts.Export()
 	if err := jopts.Validate(); err != nil {
 		err = errors.Wrap(err, "problem validating cache options")
-		return &OperationOutcome{Success: false, Text: errors.Wrap(err, "problem validating cache options").Error()}, errors.Wrap(err, "problem validating cache options")
+		return &OperationOutcome{
+			Success: false,
+			Text:    errors.Wrap(err, "problem validating cache options").Error(),
+		}, errors.Wrap(err, "problem validating cache options")
```
tychoish: should this be nil?

jrpc/internal/service.go
Utagai: Yup this is what I meant by "It may be worth it to do something similar to other times where we return (*OperationOutcome, error) and its better to send over the *OperationOutcome instead.".

I think we could just include this in this PR itself if we wanted. I'll go ahead and do this.

Ditto your other question.

PR #26, opened 2019-01-03
MAKE-569, MAKE-601: Fix TestManagerInterface/REST failures on RACE, Fix context deadline exceeded errors for JRPC process creation

Utagai:
Fundamentally, the problem here is twofold:

1. We were polluting context use by mixing up the request context with the context used by the process, which isn't a good thing especially when the process methods can hang.
2. We were calling `proc.Info()` in multiple parts of the codebase without calling `proc.Wait()`, which can lead to hangs, especially if the context used has no deadline.

It is possible that this may resolve [MAKE-523](https://jira.mongodb.org/browse/MAKE-523).

### DIFF
--- a/process_blocking.go
+++ b/process_blocking.go
@@ -184,10 +184,10 @@ func (p *blockingProcess) Info(ctx context.Context) ProcessInfo {
 		case res := <-out:
 			return res
 		case <-ctx.Done():
-			return ProcessInfo{}
+			return p.getInfo()
 		}
 	case <-ctx.Done():
-		return ProcessInfo{}
+		return p.getInfo()
 	}
 }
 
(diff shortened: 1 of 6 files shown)


process_blocking.go
```diff
@@ -184,10 +184,10 @@ func (p *blockingProcess) Info(ctx context.Context) ProcessInfo {
 		case res := <-out:
 			return res
 		case <-ctx.Done():
-			return ProcessInfo{}
+			return p.getInfo()
```
Utagai: It's worth noting that `getInfo()` will return an empty `ProcessInfo` if the process has not yet terminated and the context has reached its deadline or was canceled. This is actually not as good as we _could_ be. A better idea here is probably to change the `blockingProcess` implementation so that it holds a `cmd` object, and uses this `cmd` object to return a partially-filled-in `ProcessInfo` struct from `getInfo()`.

PR #29, opened 2019-01-08
MAKE-523: Re-introduce a better basicProcess

Utagai:
This `Process` implementation has my ideas for a good `Process` implementation merged into what we had initially on Friday, as well as a bunch of bug fixes.

### DIFF
--- a/process_basic.go
+++ b/process_basic.go
@@ -60,24 +60,17 @@ func newBasicProcess(ctx context.Context, opts *CreateOptions) (Process, error) 
 	p.info.Options = p.opts
 	p.info.Host = p.hostname
 
-	go p.reactor(ctx, cmd)
+	go p.transition(ctx, cmd)
 
 	opts.started = true
 
 	return p, nil
 }
 
-func (p *basicProcess) reactor(ctx context.Context, cmd *exec.Cmd) {
-	errs := make(chan error)
-	sig := make(chan struct{})
-	go func() {
-		defer close(errs)
-		close(sig)
-		errs <- cmd.Wait()
-	}()
-	<-sig
+func (p *basicProcess) transition(ctx context.Context, cmd *exec.Cmd) {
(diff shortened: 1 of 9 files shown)


process_basic.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	p := &basicProcess{
+		id:            id,
+		opts:          *opts,
+		cmd:           cmd,
+		tags:          make(map[string]struct{}),
+		waitProcessed: make(chan struct{}),
+		initialized:   make(chan struct{}),
+		triggers:      ProcessTriggerSequence{},
```
tychoish: I think the zero value of this is fine.

process_basic.go
```diff
@@ -0,0 +1,208 @@
...
+	waitFinished := make(chan error)
+
+	initialize := func() {
+		p.Lock()
+		defer p.Unlock()
+		p.opts.started = true
+		p.info.IsRunning = true
+		p.info.PID = p.cmd.Process.Pid
+		p.cmd = cmd
+		close(p.initialized)
```
tychoish: defer closes?

process_basic.go
```diff
@@ -0,0 +1,208 @@
...
+
+	opts.started = true
+
+	return p, nil
+}
+
+func (p *basicProcess) transition(ctx context.Context, cmd *exec.Cmd) {
+	waitFinished := make(chan error)
+
+	initialize := func() {
```
tychoish: I feel like you only call initialize once, so you might as well move the function call/body there so it  reads more realistically.

Utagai: Moved the definitions of `initialize()` and `finish()` close to their call-site, but kept the variables instead of using something like `func() { ... }()`. I think this serves as good self-documentation about what the two blocks of code do in terms of transitioning the process through phases.

PR #30, opened 2019-01-09
MAKE-537: Add some missing tests for basicProcess in JRPC & integration_test.go

Utagai:
These are some tests elsewhere in the codebase where I forgot to add `basicProcess`.

### DIFF
--- a/process_basic.go
+++ b/process_basic.go
@@ -161,9 +161,6 @@ func (p *basicProcess) Wait(ctx context.Context) (int, error) {
 	case <-p.waitProcessed:
 	}
 
-	if !p.info.Successful {
-		return p.info.ExitCode, errors.New("operation failed")
-	}
 	return p.info.ExitCode, p.err
 }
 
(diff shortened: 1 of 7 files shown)


process_basic.go
```diff
@@ -161,6 +161,9 @@ func (p *basicProcess) Wait(ctx context.Context) (int, error) {
 	case <-p.waitProcessed:
 	}
 
+	if !p.info.Successful {
```
Utagai: This is something that has to be done because of tests in `integration_test.go` assuming that unsuccessful executions will trigger this error with this message in calls to `Process.Wait()`.

Honestly, I think this is bad because it is an undocumented and very particular behavior that `Wait()` needs to conform to which doesn't add much value (it should be trivial to call `Process.Info()` after calling `Process.Wait()` and checking `ProcessInfo.Successful`). This little detail of how `Process.Wait()` should be implemented is easy to miss.

Utagai:
Yup; the two failures were because of a race that should now be fixed and a Windows-specific issue that I've updated `integration_test.go` to be more lax about and avoid. They are both green now.

PR #37, opened 2019-02-08
MAKE-638, MAKE-659: add process tracking support and pre-signal mongod shutown trigger for Windows

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-638
https://jira.mongodb.org/browse/MAKE-659
## Changes
* Write wrappers for Windows API calls for events and job objects. Some of the job object stuff is taken from Evergreen's existing subprocess.
* Write an interface for platform-specific process tracking and unit tests for it.
    * All of the implementations are stubs for now aside from Windows, will be implemented later.
* Add process tracker to `basicProcessManager`.
* Windows process tracker keeps 1 job object per manager instance with a flag to create the manager with/without a job object for the processes.
* Wrote signal trigger to cleanly shut down mongod and tests.
* Trivial spellchecking

## Some questions:
* Would it be better to put this in its own subpackage to avoid messiness?
* Should process tracking support contexts (i.e. should process tracking functions be able to time out and cancel)?
* The Windows process tracking tests are skipped in Evergreen due to the agent making its own job objects. I can try fiddling with it to make the tests pass by ensuring processes break away from the Evergreen agent job, but I don't know if this is a worthwhile investment of time. They pass on Windows spawn hosts.

### DIFF
--- a/jasper_test.go
+++ b/jasper_test.go
@@ -85,7 +85,8 @@ outerRetry:
 			return nil, -1
 		default:
 			port := getPortNumber()
-			srv := NewManagerService(NewLocalManager())
+			localManager, _ := NewLocalManager(false)
+			srv := NewManagerService(localManager)
 			app := srv.App()
 			app.SetPrefix("jasper")
 			if err := app.SetPort(port); err != nil {
--- a/jasper_test.go
+++ b/jasper_test.go
@@ -85,7 +85,10 @@ outerRetry:
 			return nil, -1
 		default:
 			port := getPortNumber()
-			localManager, _ := NewLocalManager(false)
+			localManager, err := NewLocalManager(false)
+			if err != nil {
+				return nil, -1
+			}
 			srv := NewManagerService(localManager)
 			app := srv.App()
 			app.SetPrefix("jasper")
--- a/process_basic.go
+++ b/process_basic.go
@@ -48,7 +48,9 @@ func newBasicProcess(ctx context.Context, opts *CreateOptions) (Process, error) 
 		p.Tag(t)
 	}
 
-	_ = p.RegisterTrigger(ctx, makeOptionsCloseTrigger())
+	if err = p.RegisterTrigger(ctx, makeOptionsCloseTrigger()); err != nil {
+		return nil, errors.Wrap(err, "problem registering options closer trigger")
+	}
 
 	err = cmd.Start()
 	if err != nil {
--- a/process_basic.go
+++ b/process_basic.go
@@ -48,7 +48,7 @@ func newBasicProcess(ctx context.Context, opts *CreateOptions) (Process, error) 
 		p.Tag(t)
 	}
 
-	p.RegisterTrigger(ctx, makeOptionsCloseTrigger())
+	_ = p.RegisterTrigger(ctx, makeOptionsCloseTrigger())
 
 	err = cmd.Start()
 	if err != nil {
@@ -135,8 +135,10 @@ func (p *basicProcess) Signal(_ context.Context, sig syscall.Signal) error {
 	defer p.RUnlock()
 
 	if p.Running(nil) {
-		p.signalTriggers.Run(p.Info(nil), sig)
-		return errors.Wrapf(p.cmd.Process.Signal(sig), "problem sending signal '%s' to '%s'", sig, p.id)
+		if skipSignal := p.signalTriggers.Run(p.Info(nil), sig); !skipSignal {
+			return errors.Wrapf(p.cmd.Process.Signal(sig), "problem sending signal '%s' to '%s'", sig, p.id)
+		}
+		return nil
 	}
 
 	return errors.New("cannot signal a process that has terminated")
--- a/process_blocking.go
+++ b/process_blocking.go
@@ -50,7 +50,7 @@ func newBlockingProcess(ctx context.Context, opts *CreateOptions) (Process, erro
 		p.Tag(t)
 	}
 
-	p.RegisterTrigger(ctx, makeOptionsCloseTrigger())
+	_ = p.RegisterTrigger(ctx, makeOptionsCloseTrigger())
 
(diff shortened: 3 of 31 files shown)


jasper_test.go
```diff
@@ -85,7 +85,8 @@ outerRetry:
 			return nil, -1
 		default:
 			port := getPortNumber()
-			srv := NewManagerService(NewLocalManager())
+			localManager, _ := NewLocalManager(false)
```
tychoish: what about this error?

process_basic.go
```diff
@@ -48,7 +48,7 @@ func newBasicProcess(ctx context.Context, opts *CreateOptions) (Process, error)
 		p.Tag(t)
 	}
 
-	p.RegisterTrigger(ctx, makeOptionsCloseTrigger())
+	_ = p.RegisterTrigger(ctx, makeOptionsCloseTrigger())
```
tychoish: should we error if this errors?

tracker_linux.go
```diff
@@ -0,0 +1,17 @@
+package jasper
+
+// TODO: implement
```
tychoish: should we wait for this to be done?

triggers_windows.go
```diff
@@ -0,0 +1,65 @@
...
+	// The name for the mongod termination event has the format "Global\Mongo_<pid>".
+	mongodShutdownEventNamePrefix = "Global\\Mongo_"
+	mongodShutdownEventTimeout    = 60 * time.Second
+)
+
+// makeMongodEventTrigger is necessary to support clean termination of mongods on Windows by signaling the mongod
+// shutdown event and waiting for the process to return.
+func makeMongodShutdownSignalTrigger() SignalTrigger {
+	return func(info ProcessInfo, sig syscall.Signal) (skipSignal bool) {
+		skipSignal = false
```
tychoish: you don't need to do this.

triggers_windows.go
```diff
@@ -0,0 +1,65 @@
...
+const (
+	// The name for the mongod termination event has the format "Global\Mongo_<pid>".
+	mongodShutdownEventNamePrefix = "Global\\Mongo_"
+	mongodShutdownEventTimeout    = 60 * time.Second
+)
+
+// makeMongodEventTrigger is necessary to support clean termination of mongods on Windows by signaling the mongod
+// shutdown event and waiting for the process to return.
+func makeMongodShutdownSignalTrigger() SignalTrigger {
+	return func(info ProcessInfo, sig syscall.Signal) (skipSignal bool) {
```
tychoish: I would not do the declaration here.

triggers_windows.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+
+		mongodProc, err := OpenProcess(PROCESS_ALL_ACCESS, false, uint32(pid))
+		if err != nil {
+			grip.Errorf(errors.Wrapf(err, "failed to get process handle for pid %d", pid).Error())
+			return
+		}
+		defer CloseHandle(mongodProc)
+
+		grip.Infof("mongod with pid %d received termination signal %d and was cleanly shut down", pid, sig)
```
tychoish: debug at best, prefer structured messages to formatted ones.

tracker_windows_test.go
```diff
@@ -0,0 +1,107 @@
...
+			assert.NotNil(t, cmd.ProcessState)
+			waitStatus := cmd.ProcessState.Sys().(syscall.WaitStatus)
+			assert.True(t, waitStatus.Exited())
+		},
+	} {
+		t.Run(testName, func(t *testing.T) {
+			// TODO: since processes can only be associated with 1 job, support Windows process and create them
+			// with the CREATE_BREAKAWAY_FROM_JOB flag.
+			t.Skip("Evergreen makes its own job object, so these will not pass in Evergreen tests",
+				"(although they will pass if locally run).")
```
tychoish: there's an environment variable that the evergreen agent starts, so you can check to see if you're running in this environment before just skipping.

jasper_test.go
Kimchelly: Added error check.

process_basic.go
Kimchelly: Yes it should error. I added an error check.

tracker_linux.go
Kimchelly: This should eventually be implemented when it handles cgroups.

triggers_windows.go
Kimchelly: Where would you prefer to put this declaration? As in it should go in a different file?

triggers_windows.go
Kimchelly: Removed.

triggers_windows.go
Kimchelly: I removed this since it does not seem necessary.
What exactly do you mean by structured messages?

tracker_windows_test.go
Kimchelly: Added a check for this.

Kimchelly:
Race detector failed but in an unrelated download test (this has been failing on master for a while). I made a ticket to fix it.

process_blocking.go
```diff
@@ -50,7 +50,7 @@ func newBlockingProcess(ctx context.Context, opts *CreateOptions) (Process, erro
 		p.Tag(t)
 	}
 
-	p.RegisterTrigger(ctx, makeOptionsCloseTrigger())
+	_ = p.RegisterTrigger(ctx, makeOptionsCloseTrigger())
```
tychoish: should we catch this error?

triggers_windows.go
tychoish: sorry, I would avoid declaring the skipSignal variable in the return statement, and just return false or true in each statement.

process_blocking.go
Kimchelly: Added an error check.

PR #38, opened 2019-02-15
MAKE-654: add function to register pre-signal trigger by ID for REST and RPC; always send SIGKILL to Windows instead of SIGTERM

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-654
## Changes
* Added ability to register signal trigger by unique trigger ID.
* Added tests for new functionality.
* `jasperService` returns either `OperationOutcome` or `error` but not both. This is mostly to avoid confusion, because GRPC won't return the OperationOutcome if an error is returned from `jasperService`.
* All code paths that can signal a process should send SIGKILL instead of SIGTERM to Windows processes.
* Removed some old unreachable code that checked for `ProcessInfo.ID == nil`.
* Fixed some old tests.


rest_client.go
```diff
@@ -425,6 +434,27 @@ func (p *restProcess) RegisterSignalTrigger(_ context.Context, _ SignalTrigger)
...
+		return errors.Wrap(err, "problem building request")
+	}
+
+	req = req.WithContext(ctx)
+
+	resp, err := p.client.client.Do(req)
+	if err != nil {
+		return errors.Wrap(err, "problem making request")
+	}
+	defer resp.Body.Close()
```
tychoish: should we wrap up more of lines 439-449 into a function that we could reuse more places (like handle error?) also what does handleError do?

triggers.go
```diff
@@ -45,6 +46,24 @@ func (s SignalTriggerSequence) Run(info ProcessInfo, sig syscall.Signal) (skipSi
 	return
 }
 
+// SignalTriggerID is the numerical representation of a signal trigger.
+type SignalTriggerID int
```
tychoish: maybe it makes sense to do this by string id and then have a registry (like the one in amboy for jobs/dependencies, or in evergreen for commands,) mapping them to factories rather than needing to know the integer...

also I'm not sure that iota+1 will work...

rest_client.go
Kimchelly: I replaced this with `doRequest`. Since there are a lot of places that do essentially the same thing, I changed all those places to use `doRequest`.

command_test.go
```diff
@@ -270,6 +270,7 @@ func TestCommandImplementation(t *testing.T) {
 	}
 }
 
+// TODO: fix failure due to context deadline exceeded.
```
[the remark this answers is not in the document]
Kimchelly: There's a ticket in the next sprint: https://jira.mongodb.org/browse/MAKE-666

jasper.proto
```diff
@@ -71,81 +71,80 @@ message SplunkOptions {
...
-  repeated string args = 1;
-  string working_directory = 2;
-  map<string, string> environment = 3;
-  bool override_environ = 4;
-  int64 timeout_seconds = 5;
-  repeated string tags = 6;
-  repeated CreateOptions on_success = 7;
-  repeated CreateOptions on_failure = 8;
-  repeated CreateOptions on_timeout = 9;
-  OutputOptions output = 10;
```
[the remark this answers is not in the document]
Kimchelly: I think that was vim auto-reindenting stuff. I undid the whitespace changes.

process_blocking.go
```diff
@@ -300,6 +305,14 @@ func (p *blockingProcess) RegisterSignalTrigger(_ context.Context, trigger Signa
 	return nil
 }
 
+func (p *blockingProcess) RegisterSignalTriggerID(_ context.Context, id SignalTriggerID) error {
+	trigger, err := id.MakeSignalTrigger()
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
Kimchelly: Done.

triggers.go
Kimchelly: I changed this to use a registry.

process_blocking.go
```diff
@@ -246,6 +247,10 @@ func (p *blockingProcess) Signal(ctx context.Context, sig syscall.Signal) error
 		}
 
 		if skipSignal := p.signalTriggers.Run(p.getInfo(), sig); !skipSignal {
+			// Windows does not support SIGTERM, so terminate with SIGKILL.
+			if runtime.GOOS == "windows" && sig == syscall.SIGTERM {
+				sig = syscall.SIGKILL
+			}
```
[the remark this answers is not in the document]
Kimchelly: Done.

rest_service.go
```diff
@@ -79,6 +78,7 @@ func (s *Service) App() *gimlet.APIApp {
 	app.AddRoute("/process/{id}/metrics").Version(1).Get().Handler(s.processMetrics)
 	app.AddRoute("/process/{id}/signal/{signal}").Version(1).Patch().Handler(s.signalProcess)
 	app.AddRoute("/process/{id}/logs").Version(1).Get().Handler(s.getLogs)
+	app.AddRoute("/process/{id}/signal-trigger/{trigger-id}").Version(1).Patch().Handler(s.registerSignalTriggerID)
```
tychoish: "/process/{id}/signal/{triggerID}" 

i'm sort of against using hypens (I actually like them, but they're confusing in non-lisp languages,.)

signal.go
```diff
@@ -49,18 +38,16 @@ func TerminateAll(ctx context.Context, procs []Process) error {
 	}
 
 	for _, proc := range procs {
-		proc.Wait(ctx)
+		_, _ = proc.Wait(ctx)
```
tychoish: why not catch this error?

signal.go
```diff
@@ -71,7 +58,7 @@ func KillAll(ctx context.Context, procs []Process) error {
...
```
tychoish: should we catch this error.

triggers.go
```diff
@@ -45,6 +45,25 @@ func (s SignalTriggerSequence) Run(info ProcessInfo, sig syscall.Signal) (skipSi
...
+// TODO: remove
+// MakeSignalTrigger creates a signal trigger represented by the SignalTriggerID.
+// func (id SignalTriggerID) MakeSignalTrigger() (SignalTrigger, error) {
+//     switch id {
+//     case MongodShutdownSignalTrigger:
+//         return makeMongodShutdownSignalTrigger(), nil
+//     default:
+//         return nil, errors.New("unknown signal trigger id")
+//     }
+// }
```
tychoish: should delete?

signal.go
Kimchelly: The error has never been checked on this call to `Wait()` because it will only return nil error if `(*Cmd) Wait()` runs and exits with zero exit status. This should happen on Unix systems but not on Windows. I guess I could write a platform-specific error check here for non-Windows systems.

signal.go
Kimchelly: Since this is `SIGKILL`, `Wait()` will always return an error when the process is killed.

triggers.go
Kimchelly: Removed.

rest_service.go
Kimchelly: This is already being used by `signalProcess`, so I changed this to `/process/{id}/trigger/{trigger-id}`.
Should `/process/{id}/buildlogger-urls` also be changed?

rest_service.go
```diff
@@ -78,7 +78,7 @@ func (s *Service) App() *gimlet.APIApp {
 	app.AddRoute("/process/{id}/metrics").Version(1).Get().Handler(s.processMetrics)
 	app.AddRoute("/process/{id}/signal/{signal}").Version(1).Patch().Handler(s.signalProcess)
 	app.AddRoute("/process/{id}/logs").Version(1).Get().Handler(s.getLogs)
-	app.AddRoute("/process/{id}/signal-trigger/{trigger-id}").Version(1).Patch().Handler(s.registerSignalTriggerID)
+	app.AddRoute("/process/{id}/trigger/{trigger-id}").Version(1).Patch().Handler(s.registerSignalTriggerID)
```
tychoish: I worry that this is ambiguous with post-run triggers... should we put /trigger/signal/{id} to clarify?

Kimchelly: I changed this to `/trigger/signal`.

PR #42, opened 2019-02-28
MAKE-671: add clean termination signal trigger

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-671

The webhook doesn't seem to be running for some reason, so here's a manual patch: https://evergreen.mongodb.com/version/5c7814051e2d170e99578caa

## Changes
* Added signal trigger for general clean termination of WIndows processes.

* Fix bug in Jasper service when returning error.
* Removed benchmarks from Evergreen tasks.


evergreen.yaml
```diff
@@ -79,7 +79,7 @@ tasks:
     name: race
 
   - <<: *run-build
-    tags: ["test"]
+    tags: ["bench"]
```
bsamek: I don't see any variants that run this tag.

signal.go
```diff
@@ -11,17 +11,16 @@ import (
 
 // Terminate sends a SIGTERM signal to the given process under the given
 // context. This does not guarantee that the process will actually die. This
-// function does not Wait() on the given process upon sending the signal. On
-// Windows, this function sends a SIGKILL instead of SIGTERM.
+// function does not Wait() on the given process upon sending the signal.
 func Terminate(ctx context.Context, p Process) error {
-	return errors.WithStack(p.Signal(ctx, makeCompatible(syscall.SIGTERM)))
+	return errors.WithStack(p.Signal(ctx, syscall.SIGTERM))
```
bsamek: Why did you remove `makeCompatible`?

triggers_windows.go
```diff
@@ -1,55 +1,204 @@
...
+		waitStatus, err := WaitForSingleObject(proc, mongodShutdownEventTimeout)
+		if err != nil || waitStatus != WAIT_OBJECT_0 {
+			grip.Error(message.WrapError(err, message.Fields{
+				"id":      info.ID,
+				"message": "failed to wait on process",
+				"source":  mongodShutdownSignalTriggerSource,
+			}))
+			return false
+		}
+		if waitStatus != WAIT_OBJECT_0 {
```
bsamek: This branch will never execute, because the previous branch checks the same condition and returns.

triggers_windows.go
```diff
@@ -1,55 +1,204 @@
...
+					"message": "failed to open process",
+				}))
+			}
+			return false
+		}
+		defer CloseHandle(proc)
+
+		if err := TerminateProcess(proc, 0); err != nil {
+			// TerminateProcess returns ERROR_ACCESS_DENIED if the process has already died.
+			if err != ERROR_ACCESS_DENIED {
```
bsamek: Should there be a branch where `err == ERROR_ACCESS_DENIED`? I don't follow the `err != ERROR_ACCESS_DENIED` logic here.

evergreen.yaml
Kimchelly: I added tasks to the evergreen.yaml.

signal.go
Kimchelly: makeCompatible is supposed to convert signals that don't work on certain platforms to ones that will (e.g. Windows doesn't accept SIGTERM so it gets converted to SIGKILL). However, the presignal triggers that run before actually sending the signal need to know the unmodified signal being requested. This just makes it clearer so that if somebody added a presignal trigger and expected that triggers to catch a particular signal, it would do so correctly without the signal being changed before the trigger runs. Once all the triggers run, the signal still gets modified before it calls the actual go library Signal() function.

triggers_windows.go
Kimchelly: Removed the previous branch.

triggers_windows.go
Kimchelly: This block (L170-178) is for the case when the process is known to be dead already.

If `err == ERROR_ACCESS_DENIED` the logic after this block (L180-200) has to verify that the process actually exited. This is done by checking the exit code, and if it returns `STILL_ACTIVE`, the process is still alive.

PR #45, opened 2019-03-20
MAKE-682: import oomkill detection code from evergreen

tychoish:


rest_service.go
```diff
@@ -83,26 +86,38 @@ func (s *Service) App() *gimlet.APIApp {
...
+	go s.backgroundPrune(ctx)
 
 	return app
 }
 
-func (s *Service) backgroundPrune() {
+func (s *Service) backgroundPrune(ctx context.Context) {
+	defer func() {
+		err := recovery.HandlePanicWithError(recover(), nil, "background pruning")
+		if ctx.Err() != nil || err == nil {
```
bsamek: Why return if the error is nil?

tychoish: because it's in a defer: if the context was canceled *or* if there was no panic (and we're just returning,) then we want to exit, otherwise respawn the thread

rest_service.go
```diff
@@ -68,6 +69,8 @@ func (s *Service) App() *gimlet.APIApp {
 	app.AddRoute("/download/mongodb").Version(1).Post().Handler(s.downloadMongoDB)
 	app.AddRoute("/list/{filter}").Version(1).Get().Handler(s.listProcesses)
 	app.AddRoute("/list/group/{name}").Version(1).Get().Handler(s.listGroupMembers)
+	app.AddRoute("/list/oom").Version(1).Delete().Handler(s.oomTrackerClear)
```
ablack12: Is there a reason this is also under /list? Strange route for a Delete.

subtree_oom_default.go
```diff
@@ -0,0 +1,17 @@
+// +build !linux,!darwin
+
+package tracker
```
Kimchelly: jasper

tychoish: done?

rest_service.go
tychoish: it's saying delete everything in the list? I'm not sure I can see an option that I like more though. 

I often feel like there are no right answers in REST APIs.

PR #47, opened 2019-03-21
MAKE-680: clone CreateOptions before spawning new processes with them

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-680

## Changes
* Added copy function.
* Copied all CreateOptions in any place that spawns new processes.
* Added more documentation for CreateOptions.

### DIFF
--- a/create.go
+++ b/create.go
@@ -192,9 +192,9 @@ func (opts *CreateOptions) Close() error {
 	return catcher.Resolve()
 }
 
-// Copy returns a copy of the exported fields in CreateOptions. The caller is
-// responsible for re-populating the unexported fields, which are cleared.
-func (opts *CreateOptions) CopyExported() *CreateOptions {
+// Copy returns a copy of the options for only the exported fields. Unexported
+// fields are cleared.
+func (opts *CreateOptions) Copy() *CreateOptions {
 	optsCopy := *opts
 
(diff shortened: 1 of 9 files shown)


create.go
```diff
@@ -191,3 +191,48 @@ func (opts *CreateOptions) Close() error {
 	}
 	return catcher.Resolve()
 }
+
+// Copy returns a copy of the exported fields in CreateOptions. The caller is
+// responsible for re-populating the unexported fields, which are cleared.
+func (opts *CreateOptions) CopyExported() *CreateOptions {
```
tychoish: I'd call this Copy or Export

output.go
```diff
@@ -411,3 +411,21 @@ func (o *OutputOptions) GetError() (io.Writer, error) {
 
 	return o.errorMulti, nil
 }
+
+// CopyExported returns a copy of the options for only the exported fields.
+// Unexported fields are cleared.
+func (o *OutputOptions) CopyExported() *OutputOptions {
```
tychoish: copy or export

create.go
Kimchelly: I renamed this to Copy.

output.go
Kimchelly: Copy

PR #51, opened 2019-03-24
MAKE-597: create options setter

tychoish:


command.go
```diff
@@ -116,6 +155,23 @@ func (c *Command) ContinueOnError(cont bool) *Command { c.continueOnError = cont
...
+// SuppressStandardError, when set to true, discards all standard
+// error content.
+func (c *Command) SuppressStandardError(v bool) *Command { c.opts.Output.SuppressError = v; return c }
+
+// SuppressStandardOutput, when set to true, discards all standard
+// output content.
+func (c *Command) SuppressStandardOutput(v bool) *Command { c.opts.Output.SuppressOutput = v; return c }
+
+// RedirectOutputToError sends all standard output content to standard error.
+func (c *Command) RedirectOuputToError(v bool) *Command { c.opts.Output.SendOutputToError = v; return c }
```
Kimchelly: Output*

command.go
```diff
@@ -351,9 +407,9 @@ func getCreateOpt(ctx context.Context, args []string, dir string, env map[string
 func (c *Command) getCreateOpts(ctx context.Context) ([]*CreateOptions, error) {
 	out := []*CreateOptions{}
 	catcher := grip.NewBasicCatcher()
-	if c.opts.Hostname != "" {
```
Kimchelly: I feel like there can be some confusion here due to there being a `(CreateOptions).Hostname` field and a separate `(RemoteCommandOptions).host`. `ApplyFromOpts` could be called with `(CreateOptions).Hostname` set, but if `c.remote.host` is not set to the same value by `(*Command).Host()`, it would unexpectedly ignore `(CreateOptions).Hostname`.

Can we change `(*Command).Host()` to set both `(CreateOptions).Hostname` and `(RemoteComandOptions).host`? Also set `c.remote.host` in `ApplyFromOpts` if `opts.Hostname` is set?

tychoish: I'd be more inclined to remove the hostname from CreateOpts entirely (and move it to process info? just not have it?) it's going to get overridden often by create process, and itsn't used much as is.

Kimchelly: That also works. I'm fine with removing (CreateOptions).Hostname.

PR #54, opened 2019-03-26
MAKE-697: service pruning programic configuration

tychoish:


rest_service.go
```diff
@@ -91,6 +91,27 @@ func (s *Service) App(ctx context.Context) *gimlet.APIApp {
 	return app
 }
 
+func (s *Service) ToggleCachePruning(v bool) {
+	s.cacheMutex.Lock()
+	defer s.cacheMutex.Unlock()
+
+	s.cacheOpts.Disabled = true
```
Kimchelly: Should be set to `v` instead of true? Also I feel like it should be `SetCachePruning` to be consistent with the other options

tychoish: done! good catch!

PR #55, opened 2019-03-28
MAKE-685: add cgroups to Jasper

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-685

## Changes
* Add process tracking for Linux.
* Not yet implemented for REST/RPC.


command.go
```diff
@@ -432,6 +434,10 @@ func (c *Command) getCreateOpts(ctx context.Context) ([]*CreateOptions, error) {
 		return nil, catcher.Resolve()
 	}
 
+	for _, opt := range out {
+		grip.Infof("create opt = %+v", *opt)
+	}
```
tychoish: I think you probably want to avoid this logging :)

interface.go
```diff
@@ -29,6 +29,7 @@ type Manager interface {
 	List(context.Context, Filter) ([]Process, error)
 	Group(context.Context, string) ([]Process, error)
 	Get(context.Context, string) (Process, error)
+	Limit(context.Context, interface{}) error
```
tychoish: what's supposed to be the value of the limit?

tracker.go
```diff
@@ -1,12 +1,63 @@
...
+import "github.com/pkg/errors"
+
 // processTracker provides a way to logically group processes that
 // should be managed collectively. Implementation details are
 // platform-specific since each one has its own means of managing
 // groups of processes.
 type processTracker interface {
 	// add begins tracking a process identified by its PID.
-	add(pid uint) error
+	add(pid int) error
```
tychoish: does it make sense to make these methods public?

subprocess_linux.go
```diff
@@ -0,0 +1,235 @@
...
+type LinuxRdma struct {
+	// Maximum number of HCA handles that can be opened. Default is "no limit".
+	HcaHandles *uint32 `json:"hcaHandles,omitempty"`
+	// Maximum number of HCA objects that can be created. Default is "no limit".
+	HcaObjects *uint32 `json:"hcaObjects,omitempty"`
+}
+
+// getActivePIDs returns a list of active PIDs on the system by listing the
+// contents of /proc and looking for entries that appear to be valid PIDs.
+func getActivePIDs() ([]int, error) {
```
tychoish: can't we just do this as a method on the manger which should be able to get the processes that we care about?

subprocess_linux.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+	}
+	return results, nil
+}
+
+// getEnvironmentVariables returns a map of environment variables for the
+// given pid. This function works by reading from /proc/${pid}/environ, so the
+// entries returned only reflect the values of the environment variables at the
+// time that the process was started.
+func getEnvironmentVariables(pid int) (map[string]string, error) {
```
tychoish: same, can't we look at the processes in a manager to find this out?

interface.go
```diff
@@ -29,6 +28,7 @@ type Manager interface {
...
```
tychoish: as per our conversation in person, I think it's best to hold off on this now, and then revist this method when we have a use case

subprocess_linux.go
Kimchelly: I removed all the functions that check `/proc` and instead used the  `(ProcessInfo).Options.Environment` as the source of environment variables.

subprocess_linux.go
Kimchelly: Removed this.

makefile
```diff
@@ -77,6 +77,8 @@ vendor-clean:
 	rm -rf vendor/github.com/mholt/archiver/tarlz4.go
 	rm -rf vendor/github.com/mholt/archiver/tarsz.go
 	rm -rf vendor/github.com/mholt/archiver/tarxz.go
+	rm -rf vendor/github.com/mongodb/amboy/vendor/github.com/mongodb/mongo-go-driver/internal/channel_connection.go
+	rm -rf vendor/github.com/mongodb/amboy/vendor/github.com/mongodb/mongo-go-driver/internal/results.go
+	rm -rf vendor/github.com/gogo/protobuf/protoc-gen-gogo/golden_test.go
```
tychoish: seems reasonable to delete the entire packages.  what is it about these specific files?

Kimchelly: There's nothing wrong with them per se, but they were mostly a minor annoyance because I have a git hook that warns me when files are not `gofmt`'d. I removed these lines from `vendor-clean`.

PR #59, opened 2019-04-08
MAKE-720: close file after write

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-720

### DIFF
--- a/util.go
+++ b/util.go
@@ -31,7 +31,7 @@ func makeEnclosingDirectories(path string) error {
 	return nil
 }
 
-func writeFile(reader io.Reader, path string) (err error) {
+func writeFile(reader io.Reader, path string) error {
 	if err := makeEnclosingDirectories(filepath.Dir(path)); err != nil {
 		return errors.Wrap(err, "problem making enclosing directories")
 	}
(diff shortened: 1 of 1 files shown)


util.go
```diff
@@ -30,19 +31,24 @@ func makeEnclosingDirectories(path string) error {
...
 	if err != nil {
-		return err
+		return errors.Wrap(err, "problem creating file")
 	}
 
-	if _, err := io.Copy(file, reader); err != nil {
-		return err
-	}
+	catcher := grip.NewBasicCatcher()
+	defer func() {
```
bsamek: Unless there's a strong reason to put this logic in a defer, I think it would be better to simply have it in the end of the function.

Kimchelly: I moved the closes to all the return cases.

util.go
```diff
@@ -37,12 +38,14 @@ func writeFile(reader io.Reader, path string) error {
...
 	file, err := os.Create(path)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "problem creating file")
 	}
-
 	if _, err := io.Copy(file, reader); err != nil {
-		return err
+		catcher := grip.NewBasicCatcher()
+		catcher.Add(errors.Wrap(err, "problem writing file"))
```
bsamek: Could you create the catcher before `io.Copy`. This would allow you to only have to call `file.Close` once, since you could add errors from both `io.Copy` and `file.Close` to the catcher.

Kimchelly: Done.

PR #60, opened 2019-04-23
MAKE-733: create Jasper CLI

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-733

* Implements all functions in the manager/process interfaces (extended functionality like in-memory log checking will be left for [MAKE-714](https://jira.mongodb.org/browse/MAKE-714))
* Reads from stdin, writes to stdout. 
* Support REST/RPC service communication
* No executable because I assume this Jasper command will go into the Curator CLI.
* Removed (CreateOptions).started because it's unused.
* `(*basicManager).List()` doesn't error anymore if it doesn't find any processes. This was changed because if `List()` returns error, it causes spurious errors in `(*basicManager).Close()` and forces users to determine if `List()`/`Close()` errored due to no processes or due to a genuine failure during execution.


cli/cli.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Jasper is the CLI interface to Jasper services.
+func Jasper() cli.Command {
+	return cli.Command{
+		Name:  "jasper",
+		Usage: "Jasper CLI to interact with Jasper services",
+		Before: mergeBeforeFuncs(
+			func(c *cli.Context) error {
+				port := c.Int(portFlagName)
+				minPort, maxPort := 0, 1<<16-1
```
tychoish: I think there are constants for maxint<>

cli/cli.go
```diff
@@ -0,0 +1,45 @@
...
+				return nil
+			},
+			func(c *cli.Context) error {
+				service := c.String(serviceFlagName)
+				if service != serviceREST && service != serviceRPC {
+					return errors.Errorf("service must be '%s' or '%s'", serviceREST, serviceRPC)
+				}
+				return nil
+			},
+		),
```
tychoish: doesn't this mean that we have to put flags in at this level?

cli/io.go
```diff
@@ -0,0 +1,205 @@
...
+// Validator represents an input that can be validated.
+type Validator interface {
+	Validate() error
+}
+
+// OutcomeResponse represents CLI-specific output describing if the request was
+// processed successfully and if not, the associated error message.  For other
+// responses that compose OutcomeResponse, their results are valid only if
+// Success is true.
+type OutcomeResponse struct {
```
tychoish: how viable is it to just json-seralize the rest models in the jasper package and use them rather than needing to have cli-package specific types?

remote.go
```diff
@@ -0,0 +1,25 @@
...
+
+import "context"
+
+// CloseFunc is a function used to close a service or close the client
+// connection to a service.
+type CloseFunc func() error
+
+// ConnectionCloser represents a connection that can be closed.
+type ConnectionCloser interface {
+	CloseConnection() error
```
tychoish: Why does this need to be its own interface?

PR #61, opened 2019-05-02
MAKE-756: Command should respect its CreateOptions settings 

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-756

* Local/remote commands use all options set by user in `CreateOptions` (they were ignoring all options set except working directory).
* Made a ticket for a logic bug in how output/error is handled by `Command` if redirection/suppression is used.
* Minor doc comment fixes.

### DIFF
--- a/command.go
+++ b/command.go
@@ -485,7 +485,9 @@ func (c *Command) exec(ctx context.Context, opts *CreateOptions, idx int) error 
 	addOutOp := func(msg message.Fields) message.Fields { return msg }
 	var err error
 	var newProc Process
-	if c.opts.Output.Output == nil {
+	// TODO: the logic below this is not strictly correct if, for example,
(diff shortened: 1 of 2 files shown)


command.go
```diff
@@ -99,8 +101,22 @@ func (c *Command) ProcConstructor(processConstructor ProcessConstructor) *Comman
...
+// * RedirectErrorToOutput
+// * RedirectOutputToError
+// * SetCombinedSender
+// * SetErrorSender
+// * SetErrorWriter
+// * SetOutputOptions
+// * SetOutputSender
+// * SetOutputWriter
+// * SuppressStandardError
+// * SuppressStandardOutput
```
tychoish: I would just comma seperate this list, godoc formating isn't very smart.

Kimchelly: Done.

PR #65, opened 2019-05-14
MAKE-790: add Wait to Command to check exit code

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-790

* Only valid for commands run with `Command.Run()`, not `Command.RunParallel()`.

### DIFF
--- a/command.go
+++ b/command.go
@@ -530,10 +530,10 @@ func (c *Command) exec(ctx context.Context, opts *CreateOptions, idx int) error 
 
 // Wait returns the exit code and error waiting for the underlying process to
 // complete.
-// For processes running in parallel, it only returns a zero exit code
-// if all the underlying processes return exit code zero; otherwise, it returns
-// a non-zero exit code. Similarly, it will return a non-nil error if any of
-// the underlying processes encounter an error while waiting.
+// For commands run with RunParallel, Wait only returns a zero exit code if all
+// the underlying processes return exit code zero; otherwise, it returns a
+// non-zero exit code. Similarly, it will return a non-nil error if any of the
+// underlying processes encounter an error while waiting.
 func (c *Command) Wait(ctx context.Context) (int, error) {
 	if len(c.procs) == 0 {
 		return 0, errors.New("cannot call wait on a command if no processes have started yet")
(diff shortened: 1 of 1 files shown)


command.go
```diff
@@ -33,6 +34,8 @@ type Command struct {
 	procIDs []string
 	remote  remoteCommandOptions
 	makep   ProcessConstructor
+
+	proc Process
```
tychoish: I think this doesn't work in the RunParallel case. would it be possible to pass the process around?

command.go
```diff
@@ -105,6 +110,7 @@ func (c *Command) ProcConstructor(processConstructor ProcessConstructor) *Comman
 // GetProcIDs returns an array of Process IDs associated with the sub-commands
 // being run. This method will return a nil slice until processes have actually
 // been created by the Command for execution.
+// TODO: this is not set for commands run with RunParallel.
```
tychoish: should we do this?

command.go
```diff
@@ -479,41 +485,46 @@ func (c *Command) exec(ctx context.Context, opts *CreateOptions, idx int) error
...
-
 	grip.Log(c.priority, addOutOp(msg))
+
 	return errors.WithStack(err)
 }
 
+// Wait returns the exit code and error waiting for the underlying process to
+// complete. This is undefined for Commands run with RunParallel.
+func (c *Command) Wait(ctx context.Context) (int, error) {
+	return c.proc.Wait(ctx)
```
tychoish: I'd have this call some function that returns the process and call wait on that

Kimchelly: I changed this to keep a slice of procs instead of just one to handle `Run` and `RunParallel` identically. The `Run` case should be the same as before (just wait on the one proc), and the `RunParallel` case should handle waiting on all the procs now.

command.go
```diff
@@ -479,41 +491,67 @@ func (c *Command) exec(ctx context.Context, opts *CreateOptions, idx int) error
...
 	}
-
 	grip.Log(c.priority, addOutOp(msg))
+
 	return errors.WithStack(err)
 }
 
+// Wait returns the exit code and error waiting for the underlying process to
+// complete.
+// For processes running in parallel, it only returns a zero exit code
```
tychoish: I think this is wrong because there's no parallelism here?

Kimchelly: This doc comment should still be true. I meant that Commands run with `RunParallel()` instead of `Run()` will behave this way. I clarified the doc comment to say this.

PR #68, opened 2019-05-16
MAKE-737: allow RPC to get in-memory logs from a process

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-737

* If process has in-memory logger, it gets the logs from it in chunks.


download.go
```diff
@@ -229,7 +229,9 @@ func createDownloadJobs(path string, urls <-chan string, catcher grip.Catcher) <
 func processDownloadJobs(ctx context.Context, processFile func(string) error) func(amboy.Queue) error {
 	return func(q amboy.Queue) error {
 		grip.Infof("waiting for %d download jobs to complete", q.Stats().Total)
-		_ = amboy.WaitCtxInterval(ctx, q, 1000*time.Millisecond)
+		if !amboy.WaitCtxInterval(ctx, q, 1000*time.Millisecond) {
```
tychoish: isn't this a second?

jasper.proto
```diff
@@ -189,6 +189,10 @@ message BuildloggerURLs {
     repeated string urls = 1;
 }
 
+message Logs {
```
tychoish: do we want to return more metadata?

jasper.proto
```diff
@@ -226,6 +230,7 @@ service JasperProcessManager {
   rpc ConfigureCache(CacheOptions) returns (OperationOutcome);
   rpc DownloadFile(DownloadInfo) returns (OperationOutcome);
   rpc DownloadMongoDB(MongoDBDownloadOptions) returns (OperationOutcome);
+  rpc GetLogs(JasperProcessID) returns (Logs);
```
tychoish: should this be a streaming API?

output.go
```diff
@@ -232,6 +233,27 @@ func (l *Logger) Configure() (send.Sender, error) {
 	return l.sender, nil
 }
 
+// GetInMemoryLogs gets the in-memory output logs from a process. If the process
+// has not been called with Process.Wait(), this is not guaranteed to produce
+// all the logs. This function assumes that there is exactly one in-memory
+// logger attached to this process's output.
+func GetInMemoryLogs(ctx context.Context, proc Process) ([]string, error) {
```
tychoish: does this need to be in the jasper package or can it go in the rpc package?

PR #70, opened 2019-05-18
MAKE-743: amboy job integration

tychoish:


job.go
```diff
@@ -0,0 +1,253 @@
...
+	j.Environment = env
+	j.WorkingDirectory = wd
+	j.SetID(fmt.Sprintf("%s.ext.%x", amboyJobName, sha1.Sum([]byte(cmd))))
+	return j
+}
+
+func (j *amboyJob) Run(ctx context.Context) {
+	defer j.MarkComplete()
+
+	if j.ExitCode >= 0 {
```
Kimchelly: The exit code returned can be -1 if the `Wait` below this has its context cancelled, even though the process has been created. Is this okay?

job.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func NewJobBasicForeground(opts *CreateOptions) amboy.Job {
+	j := amboyForegroundOutputJobFactory(newBasicProcess)
+	j.SetID(fmt.Sprintf("%s.basic.%x", j.Type().Name, opts.hash()))
+	j.Options = opts
+	return j
+}
+
+func (j *amboyForegroundOutputJob) Run(ctx context.Context) {
```
Kimchelly: Is it reasonable to consolidate the code in the three `Run()` methods given that they do almost the exact same thing?

job.go
```diff
@@ -0,0 +1,253 @@
...
+	"github.com/mongodb/amboy"
+	"github.com/mongodb/amboy/dependency"
+	"github.com/mongodb/amboy/job"
+	"github.com/mongodb/amboy/registry"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/level"
+	"github.com/mongodb/grip/send"
+)
+
+type amboyJob struct {
```
Kimchelly: Is there a reason that `amboyJob` has to have a bunch of duplicate `CreateOptions` and `OutputOptions` settings instead of just having a `CreateOptions` field?

command.go
```diff
@@ -156,6 +158,19 @@ func (c *Command) Priority(l level.Priority) *Command { c.priority = l; return c
 // ID sets the ID.
 func (c *Command) ID(id string) *Command { c.id = id; return c }
 
+// SetTags overrides any existing tags for a process with the
+// specified list. Tags are used to filter process with the manager.
```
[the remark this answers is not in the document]
tychoish: sure, but you could call tag before the process exists

job.go
tychoish: almost but not quite?

I'm not particularly worried about this, because most of the code has to do with output management, which is actually different for every job. Also the thing that would end up changing the most.

job.go
tychoish: There's probably no great way to handle that case?

job.go
tychoish: It makes it possible for the constructors to not return an error, and defer that to runtime, which I think makes some things easier, and I think flattening the object will make a bit more sense in the database.

PR #71, opened 2019-05-20
MAKE-790: fix RunParallel race to append to procs

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-790

* Fixes a race in `RunParallel` due to `Command.procs` being copied between Commands.

### DIFF
--- a/command.go
+++ b/command.go
@@ -29,11 +29,10 @@ type Command struct {
 	priority        level.Priority
 	runBackground   bool
 
-	cmds    [][]string
-	id      string
-	procIDs []string
-	remote  remoteCommandOptions
-	makep   ProcessConstructor
+	cmds   [][]string
+	id     string
+	remote remoteCommandOptions
+	makep  ProcessConstructor
 
 	procs []Process
 }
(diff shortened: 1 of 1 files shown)


command.go
```diff
@@ -326,7 +327,7 @@ func (c *Command) RunParallel(ctx context.Context) error {
 				}
 			}()
 			errs <- innerCmd.Run(ctx)
-			c.procs = append(c.procs, innerCmd.procs...)
+			procs <- innerCmd.procs
```
tychoish: the interactions with channels should probably all have contexts to avoid deadlocks

command.go
```diff
@@ -337,7 +338,9 @@ func (c *Command) RunParallel(ctx context.Context) error {
 			if !c.ignoreError {
 				catcher.Add(err)
 			}
+			c.procs = append(c.procs, <-procs...)
```
tychoish: it feels like this is probably a recipe for deadlocks, particularly if there's a panic

command.go
Kimchelly: I added a select statement to check if the context is done.

command.go
Kimchelly: See previous comment.

PR #75, opened 2019-05-30
MAKE-767: remove recursive read lock in Signal

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-767

* Deadlock was because a test attempted to send `SIGTERM` via `Process.Signal()` and recursively read lock the `RWMutex` in `basicProcess`.

### DIFF
--- a/manager_self_clearing.go
+++ b/manager_self_clearing.go
@@ -20,7 +20,7 @@ type selfClearingProcessManager struct {
 //
 // The self clearing process manager is not thread safe. Wrap with the
 // local process manager for multithreaded use.
-// TODO: allow local process manager to wrap other managers.
+// TODO: MAKE-803: allow local process manager to wrap other managers.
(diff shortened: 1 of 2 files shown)


manager_self_clearing.go
```diff
@@ -42,6 +43,7 @@ func NewSelfClearingProcessManager(maxProcs int, trackProcs bool) (Manager, erro
 //
 // The self clearing process manager is not thread safe. Wrap with the
 // local process manager for multithreaded use.
+// TODO: allow local process manager to wrap other managers.
```
tychoish: can we do this? or open a ticket rather than leaving a TODO in the code?

Kimchelly: I added the ticket number to the TODO. It would require some changing of local manager internals, since it currently assumes  that it's wrapping a basic manager.

PR #76, opened 2019-05-30
MAKE-801: remove unnecessary locks from basic process

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-801

* Since a process's ID never changes after first initialization, it doesn't need a lock.
* Removed the initialize step from `(*basicProcess).transition()` because the values can be set before the goroutine starts in `newBasicProcess`.

### DIFF
--- a/process_basic.go
+++ b/process_basic.go
@@ -23,7 +23,6 @@ type basicProcess struct {
 	triggers       ProcessTriggerSequence
 	signalTriggers SignalTriggerSequence
 	waitProcessed  chan struct{}
-	initialized    chan struct{}
 	sync.RWMutex
 }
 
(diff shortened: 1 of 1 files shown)


Kimchelly:
It seems like this step doesn't actually doing anything that isn't already done when setting up in `newBasicProcess`.
* `IsRunning` is already set to true in `newBasicProcess` after `cmd.Start()` returns with nil error.
* The underlying `basicProcess.cmd` field is set when `basicProcess` is first initialized.
* `PID` is available immediately after `cmd.Start()` since it does fork/exec and sets the (*exec.Cmd).Process.Pid field.

PR #79, opened 2019-06-04
MAKE-806: add sudo variants of Add/Append/Extend

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-806

### DIFF
--- a/command.go
+++ b/command.go
@@ -229,8 +229,43 @@ func (c *Command) Prerequisite(chk func() bool) *Command { c.prerequisite = chk;
 // Add adds on a sub-command.
 func (c *Command) Add(args []string) *Command { c.cmds = append(c.cmds, args); return c }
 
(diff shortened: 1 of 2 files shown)


command.go
```diff
@@ -229,9 +229,44 @@ func (c *Command) Prerequisite(chk func() bool) *Command { c.prerequisite = chk;
 // Add adds on a sub-command.
 func (c *Command) Add(args []string) *Command { c.cmds = append(c.cmds, args); return c }
 
+// SudoAdd is the same as Add but runs the sub-command with superuser
```
tychoish: I'd do this more like the SSH and less like this... so you'd have a UseSudo() method, and then add it into the command more like ssh does. 

you might also want it to SudoAs(username) so that you can do sudo -u type operations.

Kimchelly: I changed sudo to apply to all sub-commands and added `SudoAs`.

PR #81, opened 2019-06-10
MAKE-811: improve RPC TLS handling for credentials in memory and from files

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-811

This is to make Jasper RPC TLS configuration usable with evergreen and certdepot.
* RPC client/server can be created from file or from in-memory `rpc.Credentials`.
* `rpc.Credentials` can be created from PEM bytes or JSON-encoded file containing the CA cert, cert, and key.
* Test cert data generated from certdepot in `rpc/testdata`.


rpc/service.go
```diff
@@ -18,32 +18,49 @@ func AttachService(manager jasper.Manager, s *grpc.Server) error {
...
-			return nil, errors.Wrapf(err, "could not get server credentials from cert file '%s' and key file '%s'", certFile, keyFile)
+			return nil, errors.Wrap(err, "error generating TLS config from server credentials")
 		}
-		service = grpc.NewServer(grpc.Creds(creds))
-	} else {
-		service = grpc.NewServer()
+		opts = append(opts, grpc.Creds(credentials.NewTLS(tlsConf)))
 	}
 
+	service := grpc.NewServer(opts...)
```
julianedwards: Probably should add some logging interceptors here

rpc/client.go
```diff
@@ -19,31 +19,48 @@ type rpcClient struct {
...
+func NewClient(ctx context.Context, addr net.Addr, creds *Credentials) (jasper.RemoteClient, error) {
+	opts := []grpc.DialOption{grpc.WithBlock()}
+	if creds != nil {
+		tlsConf, err := creds.Resolve()
 		if err != nil {
-			return nil, errors.Wrapf(err, "could not get client credentials from cert file '%s'", certFile)
+			return nil, errors.Wrap(err, "could not resolve credentials into TLS config")
 		}
-		credsDialOpt = grpc.WithTransportCredentials(creds)
+		opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(tlsConf)))
```
julianedwards: Retry interceptors?

cli/service_rpc.go
```diff
@@ -39,22 +37,20 @@ func serviceCommandRPC(cmd string, operation serviceOperation) cli.Command {
...
-			},
-			cli.StringFlag{
-				Name:  certFilePathFlagName,
-				Usage: "the path to the key file",
+				Name:  credsFilePathFlagName,
+				Usage: "the path to the file containing the RPC service credentials",
 			},
 		},
-		Before: validatePort(portFlagName),
+		Before: mergeBeforeFuncs(
```
julianedwards: does this need to use mergeBeforeFuncs if it is just one function?

rpc/service.go
julianedwards: I think you forgot to push

rpc/client.go
julianedwards: Aviation retry logic only retires on [these errors](https://github.com/evergreen-ci/aviation/blob/master/retry.go#L120-L134). Would it still be too risky to retry? If the grpc request to spawn a host fails, can the host still have been spawned?

rpc/service.go
Kimchelly: I added it [here](https://github.com/mongodb/jasper/pull/81/files#diff-75b6eb34296ad137de7f0652ab98e00bR36)?

rpc/client.go
Kimchelly: I'm not going to do this for now because a lot of the RPCs will return `codes.Unknown` and aviation will retry on these errors when it shouldn't (and this will slow down tests a lot with unnecessary retrying). I made a [separate ticket](https://jira.mongodb.org/browse/MAKE-814) to handle request retries and fix up the RPCs to work with retrying more nicely.

rpc/service.go
julianedwards: Weird, it wasn't loading

PR #82, opened 2019-06-13
MAKE-714: jasper user command cli

tychoish:


cli/command.go
```diff
@@ -0,0 +1,295 @@
...
+					return errors.Wrap(c.Set(commandFlagName, strings.Join(c.Args().Tail(), " ")), "problem setting command")
+				}
+				return nil
+			}),
+		Action: func(c *cli.Context) error {
+			envvars := c.StringSlice(envFlagName)
+			cmds := c.StringSlice(commandFlagName)
+			useSudo := c.Bool(sudoFlagName)
+			sudoAs := c.String(sudoAsFlagName)
+			useShell := c.BoolT(noShellFlagName)
```
Kimchelly: I don't think this does what's intended - if noShell is unset, useShell is true. If `--noShell=true`, useShell is set to true. The only way to make it not use the shell is to do `--noShell=false`.
This should be `useShell := !c.BoolT(noShellFlagName)` or `noShell := c.BoolT(noShellFlagName)`.

cli/command.go
```diff
@@ -0,0 +1,295 @@
...
+				Usage: "specify to block until the process returns (subject to service timeouts), propogating the exit code from process",
+			},
+		),
+		Before: mergeBeforeFuncs(clientBefore(),
+			func(c *cli.Context) error {
+				if len(c.StringSlice(commandFlagName)) == 0 {
+					if c.NArg() == 0 {
+						return errors.New("must specify a command")
+					}
+					return errors.Wrap(c.Set(commandFlagName, strings.Join(c.Args().Tail(), " ")), "problem setting command")
```
Kimchelly: Is there a reason that you ignore the first arg here?

cli/command.go
Kimchelly: This should be either `useShell := !c.BoolT(noShellFlagName)` or `noShell := c.BoolT(noShellFlagName)`. Otherwise, the only case in which `useShell` is false will be if you set `--noShell=false`.

cli/command.go
```diff
@@ -0,0 +1,295 @@
...
+				Name:  noShellFlagName,
+				Usage: "when specified execute commands directly without shell. If multiple commands are specified they do not share a shell environment",
+			},
+			cli.StringSliceFlag{
+				Name:  tagFlagName,
+				Usage: "specify one or more tag names for the process",
+			},
+			cli.BoolFlag{
+				Name:  waitFlagName,
+				Usage: "specify to block until the process returns (subject to service timeouts), propogating the exit code from process",
```
Kimchelly: The command will block until `Run()` completes by defautl unless you specify `(*Command).Background(true)`. `Wait()` will only be guaranteed to work if it's not running in the background (i.e. what it's doing now).

cli/command.go
```diff
@@ -0,0 +1,295 @@
...
+				if sendKill {
+					return errors.WithStack(jasper.Kill(ctx, proc))
+				}
+				return errors.WithStack(jasper.Terminate(ctx, proc))
+			})
+		},
+	}
+}
+
+// KillALlCMD terminates all processes with a given tag, sending either TERM or KILL.
```
Kimchelly: KillAllCMD*

cli/command.go
```diff
@@ -0,0 +1,295 @@
...
+				Name:  "group",
+				Usage: "specify a tag to return a list of processes, overrides other options",
+			},
+		),
+		Before: clientBefore(),
+		Action: func(c *cli.Context) error {
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			filter := jasper.Filter("all")
```
Kimchelly: This should just be `jasper.All`. I think the `filter.go` needs to expliclty put `Filter` in front of the assignment for the const filters because Go is dumb.

cli/command.go
```diff
@@ -0,0 +1,295 @@
...
+	return cli.Command{
+		Name:  "kill",
+		Usage: "terminate processes",
+		Flags: append(clientFlags(),
+			cli.StringFlag{
+				Name:  groupFlagName,
+				Usage: "specify the group of process to kill",
+			},
+			cli.BoolFlag{
+				Name:  "kill",
```
Kimchelly: killFlagName

PR #83, opened 2019-06-18
MAKE-826: stream logs back to caller

tychoish:


cli/command.go
```diff
@@ -110,8 +123,42 @@ func RunCMD() cli.Command {
...
+						timer := time.NewTimer(0)
+						defer timer.Stop()
+
+						for {
+							select {
+							case <-ctx.Done():
+								grip.Notice("operation canceled")
+								return
+							case <-timer.C:
+								logLines, err := client.GetLogStream(ctx, cmdID, 1000)
```
bsamek: nit: make 1000 a constant

cli/command.go
```diff
@@ -110,8 +123,42 @@ func RunCMD() cli.Command {
...
+								if err != nil {
```
bsamek: Should this be resilient to a transient failure?

cli/utils.go
```diff
@@ -248,3 +250,7 @@ func runServices(ctx context.Context, makeServices ...func(context.Context) (jas
 	<-ctx.Done()
 	return closeAllServices(closeServices)
 }
+
+func randDur(window time.Duration) time.Duration {
+	return window + time.Duration(rand.Int63n(int64(window)))
```
bsamek: This returns a number in [n, 2n), not [n/2, 3n/2). Is that deliberate?

tychoish: yeah.

cli/command.go
tychoish: I think the clients can/do have middleware for doing retries for network errors and the like, so it's not useful doing the retries here.

cli/command.go
tychoish: ok, sorta.

PR #87, opened 2019-06-25
MAKE-834: add SSH manager that uses Jasper CLI

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-818

* Create manager that can SSH into a remote machine and use the Jasper CLI on that machine to make Jasper service requests.
* Testing is not great, but there's a [ticket](https://jira.mongodb.org/browse/MAKE-818) to mock the manager, command, and process to make this easier.


cli/ssh_manager.go
```diff
@@ -0,0 +1,153 @@
...
+	"encoding/json"
+	"io"
+
+	"github.com/mongodb/jasper"
+	"github.com/pkg/errors"
+)
+
+// sshManager uses SSH to access a remote machine's Jasper CLI, which has access to
+// methods in the Manager interface.
+type sshManager struct {
```
tychoish: re: conversation in person, this should wrap a jasper manager, so that we can track the ssh processes that we make remotely.

Kimchelly: There's still going to be an issue where the wrapped manager will just be creating `Command`s, which aren't mocked since the Command is not an interface. There's also no way to introspect on the Command implementation without a mock of it (hence the ticket I mentioned).

PR #96, opened 2019-07-19
MAKE-861: do not duplicate options tags on process creation

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-861

### DIFF
--- a/manager_basic.go
+++ b/manager_basic.go
@@ -173,8 +173,6 @@ func (m *basicProcessManager) Close(ctx context.Context) error {
 	termCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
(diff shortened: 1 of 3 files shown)


manager_basic.go
```diff
@@ -173,6 +173,8 @@ func (m *basicProcessManager) Close(ctx context.Context) error {
 	termCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
+	defer m.Clear(ctx)
```
tychoish: I worry that we don't always want to clear... Some idea of historic processes would probably be useful.

Kimchelly: I removed the clear from basic manager.

PR #99, opened 2019-07-24
MAKE-817, MAKE-840: fix various download test issues

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-817

Instead of downloading the mongo tarball and extracting it, the REST service download test just downloads a tiny test file. Hopefully this will make this test less flaky.

### DIFF
--- a/cli/utils_for_test.go
+++ b/cli/utils_for_test.go
@@ -143,6 +143,7 @@ func withMockStdout(t *testing.T, operation func(*os.File) error) error {
 // waitForRESTService waits until the REST service becomes available to serve
 // requests or the context times out.
 func waitForRESTService(ctx context.Context, t *testing.T, url string) {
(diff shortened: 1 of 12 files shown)


cli/utils_for_test.go
```diff
@@ -143,6 +143,7 @@ func withMockStdout(t *testing.T, operation func(*os.File) error) error {
 // waitForRESTService waits until the REST service becomes available to serve
 // requests or the context times out.
 func waitForRESTService(ctx context.Context, t *testing.T, url string) {
+	client := &http.Client{}
```
tychoish: can you put a sync.Pool in for this so that we can reuse clients?

Kimchelly: Done. All tests that need http clients take them from the pool.

PR #102, opened 2019-07-30
MAKE-869: suppress logs from SSH

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-869

### DIFF
--- a/cli/ssh_client.go
+++ b/cli/ssh_client.go
@@ -23,6 +23,10 @@ func NewSSHClient(remoteOpts jasper.RemoteOptions, clientOpts ClientOptions, tra
 	if err := remoteOpts.Validate(); err != nil {
 		return nil, errors.Wrap(err, "problem validating remote options")
 	}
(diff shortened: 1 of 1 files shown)


bsamek:
This seems reasonable, but I'm a little bit concerned about its effect on our ability to debug a problem. Suppose that Evergreen cannot log into hosts, are there logs that we will be able to examine to understand why?

Kimchelly:
If you can't log into the host, there's not actually much you can do. This is what is done now with all SSH commands currently running on the app servers. When you run an SSH command in an evergreen app server's in-memory Jasper manager, you won't see any output from the SSH command. Instead, it gives you the output of the command that was run on the remote host, not the output of the SSH command itself.

tychoish:
I think it's fine. I'm more worried about the possibility that we'll have some silent incompatibility with an old version of SSH, but that seems like a reasonable risk.

PR #104, opened 2019-07-30
MAKE-870: only return error from CLI over SSH Wait() if error text is non-empty

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-870

### DIFF
--- a/cli/ssh_process.go
+++ b/cli/ssh_process.go
@@ -117,7 +117,11 @@ func (p *sshProcess) Wait(ctx context.Context) (int, error) {
 		return resp.ExitCode, errors.WithStack(err)
 	}
 
-	return resp.ExitCode, errors.New(resp.Error)
+	if resp.Error != "" {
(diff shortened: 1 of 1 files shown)


cli/ssh_process.go
```diff
@@ -117,7 +117,11 @@ func (p *sshProcess) Wait(ctx context.Context) (int, error) {
 		return resp.ExitCode, errors.WithStack(err)
 	}
 
-	return resp.ExitCode, errors.New(resp.Error)
+	if resp.Error != "" {
```
bsamek: It is possible for `resp.ExitCode` to be > 0 and `resp.Error` to be the empty string?

Kimchelly: Yes, because the success case is not necessarily going to be exit code 0. Programs define their own semantics behind exit codes, so it's possible to get a non-zero exit code but the program is considered to have exit successfully. I left it to the user to read the docs on whatever program they're executing to know what exit code to expect on success/error.

PR #105, opened 2019-07-30
MAKE-803: allow arbitrary manager types to be wrapped in thread-safe manager

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-803

* Local managers can wrap any other manager.
* Remove skipDefaultTrigger since it's never used.

### DIFF
--- a/manager_basic.go
+++ b/manager_basic.go
@@ -55,7 +55,9 @@ func (m *basicProcessManager) CreateProcess(ctx context.Context, opts *CreateOpt
 		return nil, errors.Wrap(err, "problem constructing local process")
 	}
 
-	// TODO this will race because it runs later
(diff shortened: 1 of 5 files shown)


manager_basic.go
```diff
@@ -58,9 +56,7 @@ func (m *basicProcessManager) CreateProcess(ctx context.Context, opts *CreateOpt
 	}
 
 	// TODO this will race because it runs later
```
tychoish: is this still true?

Kimchelly: This is still true since CreateProcess starts the process and if it completes before it reaches this line, the default trigger won't be registered. However, there isn't really anything to do about this so I can just remove it.

PR #107, opened 2019-08-13
MAKE-824: add RemoteClient functionality to write a file

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-6566

* Write file with permissions
* Allows writing empty files (i.e. touch command)
* RPC allows streaming file in chunks (but could also be added to restClient/CLI without much difficulty).


jasper.proto
```diff
@@ -190,6 +190,12 @@ message DownloadInfo {
     ArchiveOptions archive_opts = 3;
 }
 
+message WriteFileInfo {
+    string path = 1;
+    bytes data = 2;
```
tychoish: shouldn't this be a streaming API?

remote.go
```diff
@@ -17,12 +17,13 @@ type CloseFunc func() error
...
-	GetLogStream(context.Context, string, int) (LogStream, error)
-	GetBuildloggerURLs(context.Context, string) ([]string, error)
-	SignalEvent(context.Context, string) error
+	ConfigureCache(ctx context.Context, opts CacheOptions) error
+	DownloadFile(ctx context.Context, info DownloadInfo) error
+	DownloadMongoDB(ctx context.Context, opts MongoDBDownloadOptions) error
+	GetLogStream(ctx context.Context, id string, count int) (LogStream, error)
+	GetBuildloggerURLs(ctx context.Context, id string) ([]string, error)
+	SignalEvent(ctx context.Context, name string) error
+	WriteFile(ctx context.Context, info WriteFileInfo) error
```
tychoish: arguably this should take an io.Reader

util.go
```diff
@@ -31,6 +32,37 @@ func PutHTTPClient(client *http.Client) {
...
+
+	if info.Perm == 0 {
+		info.Perm = 0666
+	}
+	return catcher.Resolve()
+}
+
+// DoWrite writes the data to the given path, creating the directory hierarchy as
+// needed.
+func (info WriteFileInfo) DoWrite() error {
```
tychoish: pointer receiver?

jasper.proto
Kimchelly: I changed the client/server to use client streaming.

remote.go
Kimchelly: I gave an option to pass either a byte slice or a reader but the reader will only be used for RPC or local machine writes.

util.go
Kimchelly: I changed it but ultimately, this doesn't really need a pointer receiver since it will only copy a few pointers + builtin types.

rpc/client.go
```diff
@@ -249,16 +251,70 @@ func (c *rpcClient) SignalEvent(ctx context.Context, name string) error {
...
 
-func (c *rpcClient) WriteFile(ctx context.Context, info jasper.WriteFileInfo) error {
-	resp, err := c.client.WriteFile(ctx, internal.ConvertWriteFileInfo(info))
+func (c *rpcClient) WriteFile(ctx context.Context, jinfo jasper.WriteFileInfo) error {
+	stream, err := c.client.WriteFile(ctx)
+	if err != nil {
+		return errors.Wrap(err, "error getting client stream to write file")
+	}
+
+	pp.Println("getting reader for serialization")
```
tychoish: omit

rpc/client.go
```diff
@@ -249,16 +251,70 @@ func (c *rpcClient) SignalEvent(ctx context.Context, name string) error {
...
+		if err != nil && err != io.EOF {
+			if closeErr := stream.CloseSend(); closeErr != nil {
+				return errors.Wrapf(closeErr, "error closing send stream after error during read: %s", err.Error())
+			}
+			return errors.Wrapf(err, "error reading from content source")
+		}
+
+		info := internal.ConvertWriteFileInfo(jinfo)
+		info.Content = buf[:n]
+		pp.Println("info content:", buf[:3])
```
tychoish: omit

rpc/internal/service.go
```diff
@@ -465,27 +466,67 @@ func (s *jasperService) SignalEvent(ctx context.Context, name *EventName) (*Oper
 	}, nil
 }
 
-func (s *jasperService) WriteFile(ctx context.Context, info *WriteFileInfo) (*OperationOutcome, error) {
-	jinfo := info.Export()
+func (s *jasperService) WriteFile(stream JasperProcessManager_WriteFileServer) error {
+	// kim: TODO: handle receiving all stream messages, first append/truncate,
+	// then append.
```
tychoish: is this still relevant?

util.go
```diff
@@ -50,17 +69,96 @@ func (info *WriteFileInfo) Validate() error {
...
+	reader, err := info.ContentReader()
+	if err != nil {
+		catcher.Wrap(file.Close(), "error closing file")
+		catcher.Wrap(err, "error getting file content as bytes")
+		return catcher.Resolve()
+	}
+
+	const mb = 1024 * 1024
+	buf := make([]byte, mb)
+	for n, err := reader.Read(buf); n > 0; n, err = reader.Read(buf) {
```
tychoish: can't we just use io.Copy? and use a bufio.Reader()?

util.go
```diff
@@ -50,17 +69,96 @@ func (info *WriteFileInfo) Validate() error {
...
+}
+
+// ContentBytes returns the contents to be written to the file as a byte slice.
+func (info *WriteFileInfo) ContentBytes() ([]byte, error) {
+	if err := info.validateContent(); err != nil {
+		return nil, errors.Wrap(err, "could not validate file content source")
+	}
+
+	if info.Reader != nil {
+		content, err := ioutil.ReadAll(info.Reader)
```
tychoish: bufio rather than readall?

Kimchelly: I changed `ContentBytes()` to be unexported and use it for reading from the reader in buffered chunks.

PR #117, opened 2019-09-13
MAKE-908: remote manager

tychoish:


manager_basic.go
```diff
@@ -16,6 +16,7 @@ type basicProcessManager struct {
 	procs              map[string]Process
 	skipDefaultTrigger bool
 	blocking           bool
+	remote             *options.Remote
```
Kimchelly: Remove?

options/create.go
```diff
@@ -137,6 +139,26 @@ func (opts *Create) Hash() hash.Hash {
...
+	}
+
+	var remoteCmd string
+
+	if opts.WorkingDirectory != "" {
+		remoteCmd += fmt.Sprintf("cd '%s' && ", opts.WorkingDirectory)
+	}
+
+	if len(env) != 0 {
+		remoteCmd += strings.Join(env, " ") + " "
```
Kimchelly: Same question as for WorkingDirectory, but this time for Environment.

options/create.go
```diff
@@ -137,6 +139,26 @@ func (opts *Create) Hash() hash.Hash {
...
 }
 
+func (opts *Create) resolveRemote(env []string) {
+	if opts.RemoteInfo == nil {
+		return
+	}
+
+	var remoteCmd string
+
+	if opts.WorkingDirectory != "" {
```
Kimchelly: Currently Create.Resolve() will set the same working directory for both the local and remote command. Is this what we want or should they be separate?

tychoish: I think that's fine. I think that if you're in a remote manager than you should be able to set the local directory just the same as you can locally.

options/create.go
tychoish: I think it's fine...

Kimchelly: This will cause the ssh command to fail unless both the local filesystem and the remote one have this directory (e.g. if you want to use `/home/user` as the remote working directory, `/home/user` will have to exist on the local machine as well).

Also, if we're going to keep this overloaded `WorkingDirectory`, you have to modify `Create.Validate()` to not stat the WorkingDirectory in the remote case.

options/create.go
Kimchelly: This will cause the ssh command to fail unless both the local filesystem and the remote one have this directory (e.g. if you want to use `/home/user` as the remote working directory, `/home/user` will have to exist on the local machine as well).

Also, if we're going to keep this overloaded `WorkingDirectory`, you have to modify `Create.Validate()` to not stat the WorkingDirectory in the remote case.

PR #121, opened 2019-10-02
MAKE-956: integrate timber into jasper

julianedwards:


glide.yaml
```diff
@@ -15,3 +15,4 @@ import:
   - package: github.com/evergreen-ci/aviation
   - package: github.com/tychoish/mongorpc
   - package: github.com/mongodb/ftdc
+  - package: github.com/evergreen-ci/timber
```
tychoish: I think we're double vendoring mongo-go-driver

options/log.go
```diff
@@ -215,6 +218,11 @@ func (l *Logger) Configure() (send.Sender, error) {
 		if err != nil {
 			return nil, err
 		}
+	case LogBuildloggerV3:
+		sender, err = timber.NewLogger(context.Background(), DefaultLogName, l.Options.Level, &l.Options.BuildloggerV3Options)
```
tychoish: discussed minor change and a new constructor.

PR #126, opened 2019-10-22
MAKE-942: port jasper benchmarks to poplar

julianedwards:


harness.go
```diff
@@ -0,0 +1,182 @@
...
+	res, err := s.Run(ctx, prefix)
+	if err != nil {
+		resultText = fmt.Sprintf("--- FAIL: %s\n", err)
+	} else {
+		resultText = fmt.Sprintf("--- PASS: %s\n", res.Report())
+	}
+
+	_, err = resultFile.WriteString(resultText)
+	if err != nil {
+		return errors.Wrap(err, "failed to write benchmark results to file")
```
Kimchelly: This should still close the file in case it errors while writing.

PR #129, opened 2019-10-22
MAKE-803: wrap arbitrary manager types in thread-safe manager

Kimchelly:
This was already approved [here](https://github.com/mongodb/jasper/pull/105). I have no idea why I closed the original PR without merging.


bsamek:
@Kimchelly Then just merge :-)

PR #131, opened 2019-10-25
MAKE-898: implement SSH using library

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-898

This is not really tested due to difficulty running SSH tests on evergreen. I did some manual tests to verify that at the bare minimum it can SSH and run processes.

* Vendor golang.org/x/crypto.
* Add Executor interface to allow local (i.e. `exec.Cmd`) vs. SSH behavior.
    * The remote PID is not easy to get from SSH so it always returns -1.
* Minor behavioral changes because the SSH command now runs more like a proper shell script. Environment variables and working directory are more robust.
* Arbitrary SSH binary arguments are not allowed (e.g. you can't just pass `ConnectTimeout=10` to the SSH binary). Instead, `options.Remote` has programmatic support for such options, including SSHing via a proxy.
* Trivial variable renaming stuff.


tychoish:
I would prefer if using the SSH binary were still an option and using this as an option?

Kimchelly:
Done. The default behavior is to use the old SSH behavior unless the manager is explicitly constructed to use the SSH library or the process is being created explicitly with `UseSSHLibrary` set to true.

makefile
```diff
@@ -2,10 +2,10 @@ name := jasper
 buildDir := build
 srcFiles := $(shell find . -name "*.go" -not -path "./$(buildDir)/*" -not -name "*_test.go" -not -path "*\#*")
 testFiles := $(shell find . -name "*.go" -not -path "./$(buildDir)/*" -not -path "*\#*")
-packages := $(name) cli rpc rest options mock internal/executor
+packages := $(name) cli rpc rest options mock testutil internal-executor
```
tychoish: does this mean that we end up running tests on testutil?

Kimchelly: No, it only gets used for linting because `test-testutil` is not a task in evergreen.yaml.

PR #134, opened 2019-11-04
MAKE-862: implement wire protocol service

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-862

* Implement service for responding to MongoDB wire protocol messages.
* Implement client for testing purposes.
* Fix various minor issues with makefile.

- [x] ~Waiting on [MAKE-994](https://jira.mongodb.org/browse/MAKE-994) and [MAKE-996](https://jira.mongodb.org/browse/MAKE-996) to be merged.~

PR #137, opened 2019-11-05
MAKE-862: implement wire protocol service

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-862

* Define request/response types in `wire/io.go`.
* Implement wire protocol service with manager and process RPC handlers.

Still need to implement this with support for OP_MSG in case the shell is version 4.2 or higher.


wire/io.go
```diff
@@ -0,0 +1,846 @@
...
+func intOK(ok bool) int {
+	if ok {
+		return 1
+	}
+	return 0
+}
+
+// ErrorResponse represents a response indicating whether the operation was okay
+// and errors, if any.
+type ErrorResponse struct {
```
tychoish: seems like most of the reponse building code should end up in mrpc)

Kimchelly: I don't think this makes sense given that all of the payloads are Jasper-specific and not defined in the mongo wire protocol spec.

Kimchelly:
#134 is old, do not use.

Kimchelly:
I updated this to use the shell service from mrpc.

wire/service.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/evergreen-ci/mrpc"
+	"github.com/evergreen-ci/mrpc/mongowire"
+	"github.com/evergreen-ci/mrpc/shell"
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/recovery"
+	"github.com/mongodb/jasper"
+	"github.com/pkg/errors"
+)
+
+// TODO: support jasper.RemoteClient functionality
```
tychoish: is there a ticket for this?

wire/service.go
```diff
@@ -0,0 +1,91 @@
...
+type service struct {
```
tychoish: are we worried about thread safety or would a synchronized manager be enough?

wire/service.go
Kimchelly: https://jira.mongodb.org/browse/MAKE-1000

wire/service.go
Kimchelly: The thread safety guarantees for this is the same as for the REST/RPC managers. I'm not particularly concerned about having multiple connections given that the synchronized manager already locks all the manager methods.

PR #142, opened 2019-11-19
MAKE-1028: scripting environment

tychoish:
this is basically done; I'm a bit worried about interactions with remote managers, which is currently disabled but I think might not be.

also need to implement methods for rpc and what not.


tychoish:
I feel pretty good about the current state of this now, I've added rest and RPC. I think I'll need help to get the ssh/shell/etc methods/helpers

options/scripting_python.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// NewPythonScriptingEnvironmnet generates a ScriptingEnvironment
+// taking the arguments given for later use. Use this function for
+// simple cases when you do not need or want to set as many aspects of
+// the environment configuration.
+func NewPythonScriptingEnvironmnet(path, reqtxt string, packages ...string) ScriptingEnvironment {
+	return &ScriptingPython{
+		CachedDuration:        time.Hour,
+		HostPythonInterpreter: "python3",
```
bsamek: What about the legacy python case?

options/scripting_python.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	if opts.HostPythonInterpreter == "" {
+		opts.HostPythonInterpreter = "python3"
+	}
+
+	return nil
+}
+
+func (opts *ScriptingPython) ID() string {
```
bsamek: This is potentially racy. Someone else could call ID and write to the cache. Is that a concern?

scripting_python.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (e *pythonEnvironment) Run(ctx context.Context, args []string) error {
+	return e.manager.CreateCommand(ctx).Add(append([]string{e.opts.Interpreter()}, args...)).Run(ctx)
+}
+
+func (e *pythonEnvironment) RunScript(ctx context.Context, script string) error {
+	scriptChecksum := fmt.Sprintf("%x", sha1.Sum([]byte(script)))
+	wo := options.WriteFile{
+		Path:    filepath.Join(e.opts.VirtualEnvPath, "tmp", strings.Join([]string{e.manager.ID(), scriptChecksum}, "-")+".lisp"),
```
bsamek: copy pasta lisp in python file

scripting_roswell.go
```diff
@@ -0,0 +1,87 @@
...
+func (e *roswellEnvironment) Build(ctx context.Context, dir string, args []string) error {
+	return e.manager.CreateCommand(ctx).Directory(dir).Environment(e.opts.Environment).AddEnv("ROSWELL_HOME", e.opts.Path).
+		SetOutputOptions(e.opts.Output).Add(append([]string{e.opts.Interpreter(), "dump", "executable"}, args...)).Run(ctx)
+}
+
+func (e *roswellEnvironment) Cleanup(ctx context.Context) error {
+	switch mgr := e.manager.(type) {
+	case RemoteClient:
+		return errors.Wrapf(mgr.CreateCommand(ctx).SetOutputOptions(e.opts.Output).AppendArgs("rm", "-rf", e.opts.Path).Run(ctx),
+			"problem removing remote python environment '%s'", e.opts.Path)
```
bsamek: python things in lisp file

options/scripting_python.go
tychoish: the constructors are only for default situations. you can always construct directly.

options/scripting_python.go
tychoish: yeahhhhhhhhh, I'm not super worried about it. 

I think the hashes will be stable. I'm more worried about someone having an options struct, creating it, losing track of the hash, and then changing it, and creating a different Environment, which I don't think will be a problem in reasonable situations.

PR #146, opened 2019-12-02
MAKE-1000: implement wire remote client

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1000

* Wire protocol remote client (mostly for testing purposes).
* Copy RPC manager/process tests to wire package.

### DIFF
--- a/wire/io.go
+++ b/wire/io.go
@@ -6,8 +6,6 @@ import (
 	"github.com/mongodb/jasper/options"
 )
 
-// kim: TODO: remove make* functions
-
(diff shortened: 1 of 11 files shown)


wire/io.go
```diff
@@ -6,21 +6,19 @@ import (
 	"github.com/mongodb/jasper/options"
 )
 
+// kim: TODO: remove make* functions
```
tychoish: ?

Kimchelly: This is done, so I removed it.

PR #148, opened 2019-12-02
MAKE-1043: move benchmarks out of top level package

tychoish:


options/create.go
```diff
@@ -42,6 +47,9 @@ type Create struct {
 	StandardInput      io.Reader `bson:"-" json:"-" yaml:"-"`
 	StandardInputBytes []byte    `bson:"stdin_bytes" json:"stdin_bytes" yaml:"stdin_bytes"`
 
+	Unsynchronized bool   `bson:"unsynchronized" json:"unsynchronized" yaml:"unsynchronized"`
```
Kimchelly: This is somewhat deceptive since it seems to imply that an external user who sets this or Implementation will actually change how a process is created via a manager. If we're just going to make a `NewProcess()` method, shouldn't we just export `newBasicProcess`, `newBlockingProcess`, and create a `newLocalProcess` wrapper instead since the user already has to know the implementations? Calling `NewProcess()` which devolves into a switch statement seems less clean.

tychoish: Would it be reasonable to make the managers use new process? and not have process type be controlled so strongly by process manager?

Kimchelly: It wouldn't be great support given that we have to keep synchronized manager with the guarantee that the processes it returns are also thread-safe. If you were to specify it to be unsynchronized, you wouldn't actually get rid of the thread safety on the process. I think it's not a big issue that the managers determine the type of processes created.

tychoish: hrm

PR #150, opened 2019-12-03
MAKE-1043: improved scripting harness (part 1)

tychoish:
This has: 

- support in the cli harnesses.
- add a return value from Build in the scripting method to get the path of the artifact produced. 

When this, and #148 we can move to merge all of the remote interfaces into a "remote" package to reduce test duplication, and can pull the scripting interface further out of main jasper package, but I want to hold of on that until, just to avoid extra complication.


cli/ssh_client.go
```diff
@@ -89,12 +94,50 @@ func (c *sshClient) CreateCommand(ctx context.Context) *jasper.Command {
...
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	output, err := c.runManagerCommand(ctx, CreateScriptingCommand, cliOpts)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	_, err = ExtractInfoResponse(output)
```
Kimchelly: This should be ExtractIDResponse.

Can we test this method?

cli/manager.go
```diff
@@ -75,6 +77,29 @@ func managerCreateProcess() cli.Command {
...
+			opts := &ScriptingOptions{}
+			return doPassthroughInputOutput(c, opts, func(ctx context.Context, client jasper.RemoteClient) interface{} {
+				harnessOpts, err := opts.Export()
+				if err != nil {
+					return &IDResponse{ID: client.ID(), OutcomeResponse: *makeOutcomeResponse(errors.Wrapf(err, "error creating process"))}
+				}
+
+				env, err := client.CreateScripting(ctx, harnessOpts)
+				if err != nil {
+					return &IDResponse{ID: harnessOpts.ID(), OutcomeResponse: *makeOutcomeResponse(errors.Wrapf(err, "error creating process"))}
```
Kimchelly: "error creating scripting harness"

cli/ssh_client.go
```diff
@@ -16,6 +17,10 @@ import (
 type sshClient struct {
 	manager jasper.Manager
 	opts    sshClientOptions
+	seCache struct {
```
Kimchelly: Should the usage of `se` be changed to `sh`? This is applied inconsistently throughout the code.

jasper.proto
```diff
@@ -279,6 +279,12 @@ message ScriptingEnvironmentBuildArgs {
   repeated string args = 3; 
 }
 
+message ScriptingEnvironmentBuildResponse {
```
Kimchelly: Are we keeping this scripting environment or should it be harness?

mock/manager.go
```diff
@@ -25,7 +25,7 @@ type Manager struct {
 	CreateConfig        Process
 	ManagerID           string
 	Procs               []jasper.Process
-	ScriptingEnv        jasper.ScriptingEnvironment
+	ScriptingEnv        jasper.ScriptingHarness
```
Kimchelly: Same question about harness.

util.go
```diff
@@ -1 +1,9 @@
 package jasper
+
+import "bytes"
+
+type writeCloser struct {
```
Kimchelly: This has already been defined in `command_test.go` so we should get rid of the duplicate bytes.Buffer wrapping.

rpc/internal/service.go
```diff
@@ -684,28 +684,31 @@ func (s *jasperService) ScriptingEnvironmentRunScript(ctx context.Context, args
...
-		ExitCode: 0,
-	}, nil
+		return &ScriptingHarnessBuildResponse{
+			Outcome: &OperationOutcome{
+				Success:  false,
+				Text:     err.Error(),
+				ExitCode: 1,
+			}}, nil
+	}
+	fmt.Print(path)
```
Kimchelly: Remove print.

PR #151, opened 2019-12-04
MAKE-1038: use golangci-lint

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1038

* Use golangci-lint instead of gometalinter.
* Simplify go environment setup in makefile.
* Simplify go scripting test to not depend on run-linter while still verifying that scripting packages work.


makefile
```diff
@@ -22,13 +22,18 @@ ifeq ($(OS),Windows_NT)
 gocache := $(shell cygpath -m $(gocache))
 gopath := $(shell cygpath -m $(gopath))
 endif
-goEnv := GOPATH=$(gopath) GOCACHE=$(gocache)$(if $(GO_BIN_PATH), PATH="$(shell dirname $(GO_BIN_PATH)):$(PATH)")
+
+export GOPATH := $(gopath)
+export GOCACHE := $(gocache)
+ifneq ($(GO_BIN_PATH),)
+	export PATH := $(shell dirname $(GO_BIN_PATH)):$(shell printenv PATH)
+endif
```
tychoish: I'd rather not export, it makes it easier to see what happens by copying the (admitedly too long) shell line.

Kimchelly: I actually would rather export because I've found people are highly prone to not editing the makefile correctly. Whenever someone adds something, it's too easy to forget it or accidentally rely on the go environment without realizing it (e.g. gometalinter always relied on the "go" binary in the PATH), so global state that applies to the entire makefile seems preferable in this case.

Regardless, I re-added the goEnv if you think its' easier. However, I don't trust our ability to write/edit makefiles correctly since not everyone knows how the makefiles work.

PR #152, opened 2019-12-04
MAKE-1045: better test wire protocol with different wire things

tychoish:


wire/client_test.go
```diff
@@ -290,22 +338,42 @@ func TestWireProcess(t *testing.T) {
...
+		Options testutil.OptsModify
+	}{
+		{
+			Name: "Blocking",
+			Options: func(opts *options.Create) {
+				opts.Implementation = options.ProcessImplementationBlocking
+			},
+		},
+		{
+			Name: "Blocking",
```
Kimchelly: Duplicate blocking - there should be a basic case.

wire/client_test.go
```diff
@@ -290,22 +338,42 @@ func TestWireProcess(t *testing.T) {
...
+			Options: func(opts *options.Create) {
+				opts.Implementation = options.ProcessImplementationBasic
+			},
+		},
+		{
+			Name:    "Default",
+			Options: func(opts *options.Create) {},
```
Kimchelly: This is unnecessary.

wire/client_test.go
```diff
@@ -24,17 +24,37 @@ func TestWireManager(t *testing.T) {
...
+		Options testutil.OptsModify
+	}{
+		{
+			Name: "Blocking",
+			Options: func(opts *options.Create) {
+				opts.Implementation = options.ProcessImplementationBlocking
+			},
+		},
+		{
+			Name: "Blocking",
```
Kimchelly: Copy paste.

wire/client_test.go
```diff
@@ -24,17 +24,37 @@ func TestWireManager(t *testing.T) {
...
+			Options: func(opts *options.Create) {
+				opts.Implementation = options.ProcessImplementationBasic
+			},
+		},
+		{
+			Name:    "Default",
```
Kimchelly: Copy paste.

wire/client_test.go
tychoish: I know, but making sure that you don't need to set implementation is probably a good thing?

tychoish:
it's my intention to merge rpc/rest/cli/wire into a single "remote" package (with jasper.Remote) and refactor the tests to use a single set of cases, and I'd like to do that as part of a seperate commit, as part of MAKE-1043, using this model.

wire/client_test.go
Kimchelly: This should be left to validation in create options.

tychoish: right but if we're always setting it, then we don't know if the default operation works in this configuration.

Kimchelly: I guess?

PR #154, opened 2019-12-04
MAKE-1049: use synchronized instead of unsynchronized

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1049

### DIFF
--- a/manager_synchronized.go
+++ b/manager_synchronized.go
@@ -97,6 +97,7 @@ func (m *synchronizedProcessManager) List(ctx context.Context, f options.Filter)
 
 	procs, err := m.manager.List(ctx, f)
 	syncedProcs := make([]Process, 0, len(procs))
+	var syncedProcs []Process
 	for _, proc := range procs {
 		syncedProcs = append(syncedProcs, &synchronizedProcess{proc: proc})
 	}
@@ -108,6 +109,9 @@ func (m *synchronizedProcessManager) Get(ctx context.Context, id string) (Proces
 	defer m.mu.RUnlock()
 
 	proc, err := m.manager.Get(ctx, id)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
 	return &synchronizedProcess{proc: proc}, errors.WithStack(err)
 }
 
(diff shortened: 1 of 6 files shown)


manager_synchronized.go
```diff
@@ -43,8 +43,6 @@ func (m *synchronizedProcessManager) ID() string {
 }
 
 func (m *synchronizedProcessManager) CreateProcess(ctx context.Context, opts *options.Create) (Process, error) {
-	opts.Synchronized = true
```
tychoish: why not leave this here, and then unwrap all of the returned processes? 

I think it's a bit undesireable to return the same process in multiple Get() calls, but have them be protected by different locks.

Kimchelly: Because that causes the original issue with being unable to call multiple methods in different goroutines on a single process. Requesting different copies of the same process (i.e. what's done in all the remote services) has to produce different synchronized processes.

manager_synchronized.go
```diff
@@ -53,6 +51,8 @@ func (m *synchronizedProcessManager) CreateProcess(ctx context.Context, opts *op
 		return nil, errors.WithStack(err)
 	}
 
+	opts.Synchronized = true
```
tychoish: I think modifying this here, after the process is created is probably wrong.

tychoish: I think we can just delete this line, and then merge it later, and revisit later...

manager_synchronized.go
tychoish: This feels unexpected. Let's take the modification of synchronized.True out

manager_synchronized.go
Kimchelly: Putting it before is going to break any possibility of having multithreading in remote services. If you do this, all remote service which must call Get() will use the exact same synchronized process and will be forced to do entirely single-threaded operations on processes.

tychoish: I think that's right though? On the other hand you end up with a situation where you say "give me a locked thing" you're asking for singally-threaded-access to a process?

Kimchelly: It's not though, because the assumption has always been that you can perform multiple operations on a given process. If somebody was waiting on the process in one goroutine, it doesn't make sense that the wait would block all other operations from being processed (e.g. you can't get process info, you can't ask it if it's done, all you can do is wait for it to finish). All the remote services use the synchronized manager so this would be bad as far as usability of remote services.

Kimchelly: It also has implications for potentially blocking all manager functionality which calls any method on a process. If the processes are all stuck waiting for other operations to finish which could take a long time (e.g. a wait or a signal), all your process-related requests might be locked out (e.g. a manager won't be able to close because it's blocked trying to signal processes to exit).

PR #155, opened 2019-12-05
MAKE-1036: run lint and test in task groups

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1036

This is mostly a copy-paste from cedar.

Lint is actually slower in a task group as opposed to just running a single "lint" task. The current lint task takes about 1 second per package (minus the top-level package). This new lint task group takes approximately 20-30 seconds per package.


bsamek:
> Lint is actually slower in a task group as opposed to just running a single "lint" task. The current lint task takes about 1 second per package (minus the top-level package). This new lint task group takes approximately 20-30 seconds per package.

Do you know why that is?

Kimchelly:
Most likely it's the overhead of the agent reinitializing each package linter as a new task as opposed to just running everything in a single task.

tychoish:
I bet it's doing git.get_project before every task rather than once in group_setup

Kimchelly:
The task groups still aren't as fast as just running a single lint task (100 seconds vs 60 seconds), but I guess it's not that important?

PR #162, opened 2020-02-09
MAKE-1134: support sigabort

tychoish:


signal_windows.go
```diff
@@ -3,8 +3,12 @@ package jasper
 import "syscall"
 
 func makeCompatible(sig syscall.Signal) syscall.Signal {
-	if sig == syscall.SIGTERM {
+	switch sig {
+	case syscall.SIGTERM:
 		return syscall.SIGKILL
+	case syscall.SIGABRT:
```
Kimchelly: Can we write a test for this (the same one as `WaitGivesProperExitCodeOnSignalDeath` in `process_test.go` except `SIGABRT` instead of `SIGTERM`)?

PR #169, opened 2020-03-24
MAKE-1142: cached loggers

tychoish:


options/logging.go
```diff
@@ -0,0 +1,221 @@
...
+			return nil, errors.Wrap(err, "problem parsing bson from message body")
+		}
+
+		if lp.AddMetadata {
+			return message.NewFields(lp.Priority, payload), nil
+		}
+
+		return message.NewSimpleFields(lp.Priority, payload), nil
+	default: // includes string case.
+		if lp.AddMetadata {
```
Kimchelly: This should be if `!lp.AddMetadata` shouldn't it?

options/logging.go
```diff
@@ -0,0 +1,221 @@
...
+			return message.NewSimpleBytesMessage(lp.Priority, data), nil
+		}
+
+		return message.NewBytesMessage(lp.Priority, data), nil
+	}
+}
+
```
Kimchelly: Rather than pass a bool that immediately returns for a non-bson case, the use of splitBSONSlice on L100 should be split into two different functions since they have entirely different behavior and allowing non-bson to be passed to a function that has the word "BSON" in it sounds deceptive.

options/logging.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/bson"
+)
+
+// CachedLogger is the cached item representing a processes normal
+// output. It captures information about the cached item, as well as
+// go interfaces for sending log messages.
+type CachedLogger struct {
+	ID       string    `bson:"id" json:"id" yaml:"id"`
+	Manager  string    `bson:"manager_id" json:"manager_id" yaml:"manager_id"`
```
Kimchelly: Where is this used?

tychoish: logging.

remote/internal/service.go
```diff
@@ -745,3 +746,104 @@ func (s *jasperService) ScriptingHarnessTest(ctx context.Context, args *Scriptin
...
+	if out == nil {
+		return &LoggingCacheInstance{
+			Outcome: &OperationOutcome{
+				Success: false,
+				Text:    "not found",
+			},
+		}, nil
+	}
+
+	return ConvertCachedLogger(out), nil
```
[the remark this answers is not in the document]
tychoish: ConvertCachedLogger produces the appropriate Outcome value.

remote/internal/service.go
```diff
@@ -745,3 +746,104 @@ func (s *jasperService) ScriptingHarnessTest(ctx context.Context, args *Scriptin
...
+
+	return &OperationOutcome{
+		Success: true,
+	}, nil
+}
+
+func (s *jasperService) LoggingCachePrune(ctx context.Context, arg *timestamp.Timestamp) (*OperationOutcome, error) {
+	lc := s.manager.LoggingCache(ctx)
+	if lc == nil {
+		return nil, errors.New("logging cache not supported")
```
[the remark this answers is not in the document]
tychoish: None of the other logging cache methods return OperationOutcome=false, if the logging cache is nil (which is to say, that the logging cache isn't implemented in the manager interface (which I think is equivalent to an HTTP 500, which is how I think about returning errors from service implementations. It's fine to change, but I'm not sure.

PR #173, opened 2020-04-21
MAKE-1267: add hooks

tychoish:


options/command.go
```diff
@@ -35,3 +39,88 @@ func (opts *Command) Validate() error {
...
+		}
+	}
+}
+
+// CommandPostHook describes a common function type to run after the
+// command completes and processes its error returns, potentially
+// overriding the output of the command.
+type CommandPostHook func(error) error
+
+// MergePostHooks runs all passed post hooks, collecting
```
Kimchelly, line 87: Removed "passed".

options/create.go
```diff
@@ -37,6 +37,7 @@ const (
 // execution configuration, post-execution triggers, and output configuration.
 // It is not safe for concurrent access.
 type Create struct {
+	ID               string            `bson:"id" json:"id" yaml:"id"`
```
Kimchelly, line 40: I don't like the idea of a CreateOptions with an ID. What's the difference between a `Process.ID()` versus a `CreateOptions.ID`? Why does it even need an ID? It's for specifying options, and this seems to be a field which is externally injected but does nothing.

It's also deceptive - setting this for CreateOptions does nothing to a process ID - it only affects Command. This is a weird and unexpected decision and nobody will know how to use this field. I think it adds unnecessary complexity and is a dangerous road to go down to attach IDs everywhere. It's weird to mix CreateOptions, which are for processes, and Command when Process should be independent of Command.

This isn't worth adding just so you can an ID have in a log message from Command. Log the ID for the Command object instead.

command.go
```diff
@@ -485,9 +481,25 @@ func (c *Command) AppendArgsWhen(cond bool, args ...string) *Command {
...
-func (c *Command) SetHook(h func(error) error) *Command { c.opts.Hook = h; return c }
+// Prerequisite sets a function on the Command such that the Command will only
+// execute if the function returns true. The Prerequisite function runs once per
+// Command object regardless of how many subcommands are
+// present. Prerequsite functions run even if the command's
+// RunFunction is set.
+func (c *Command) Prerequisite(chk func() bool) *Command { c.opts.Prerequisite = chk; return c }
+
+// PostHook allows you to add a function that runs (locally)
+// after the each subcommand in the Command completes. When specified
```
Kimchelly, line 492: When specified,

tychoish:
made changes for all comments, will add tests in a bit.

options/command.go
```diff
@@ -35,3 +39,89 @@ func (opts *Command) Validate() error {
...
+		catcher := grip.NewBasicCatcher()
+		for _, fn := range fns {
+			catcher.Add(fn(err))
+		}
+		return catcher.Resolve()
+	}
+}
+
+// MergeAbortingPostHook produces a post hook that returns the first
+// error returned by one of it's argument functions, and returns nil
```
Kimchelly, line 101: its

options/command_test.go
```diff
@@ -0,0 +1,147 @@
...
+		t.Run("Merged", func(t *testing.T) {
+			mock := &testPostHook{abort: true}
+			hook := MergePostHooks(mock.hook, mock.hook)
+			assert.NoError(t, hook(nil))
+			assert.Equal(t, 2, mock.count)
+			assert.Error(t, hook(errors.New("hi")))
+			assert.Equal(t, 4, mock.count)
+			assert.Equal(t, 2, mock.errsSeen)
+		})
+		t.Run("ShortCircut", func(t *testing.T) {
```
Kimchelly, line 119: ShortCircuit*

options/command.go
Kimchelly: Still need to fix this.

PR #175, opened 2020-04-22
MAKE-1197: add mrpc implementation support for scripting.harness interface

julianedwards:


remote/mdb_io.go
```diff
@@ -283,3 +284,88 @@ type loggingCacheLenRequest struct {
 type loggingSendMessageRequest struct {
 	Payload options.LoggingPayload `bson:"send_message"`
 }
+
+type scriptingCreateRequest struct {
+	Params struct {
+		Type    string `bson:"type"`
+		Options []byte `bson:"options"`
```
Kimchelly, line 291: Can't this just store the `options.ScriptingHarness` interface itself rather than send as bytes? `shell.RequestToMessage` already calls `bson.Marshal`.

remote/mdb_client.go
```diff
@@ -87,12 +88,58 @@ func (c *mdbClient) CreateCommand(ctx context.Context) *jasper.Command {
 	return jasper.NewCommand().ProcConstructor(c.CreateProcess)
 }
 
-func (c *mdbClient) CreateScripting(_ context.Context, _ options.ScriptingHarness) (scripting.Harness, error) {
-	return nil, errors.New("scripting environment is not supported")
+func (c *mdbClient) CreateScripting(ctx context.Context, opts options.ScriptingHarness) (scripting.Harness, error) {
+	marshalledOpts, err := bson.Marshal(opts)
```
Kimchelly, line 92: See my question about `scriptingCreateRequest`.

remote/mdb_service_scripting.go
```diff
@@ -18,11 +23,198 @@ const (
...
+
+func (s *mdbService) scriptingCreate(ctx context.Context, w io.Writer, msg mongowire.Message) {
+	req := &scriptingCreateRequest{}
+	if !s.serviceScriptingRequest(ctx, w, msg, req, ScriptingCreateCommand) {
+		return
+	}
+
+	opts, err := options.NewScriptingHarness(req.Params.Type)
+	if err != nil {
+		shell.WriteErrorResponse(ctx, w, mongowire.OP_REPLY, errors.Wrap(err, "problem creating harness options"), ScriptingCreateCommand)
```
Kimchelly, line 48: Return on error.

remote/mdb_service_scripting.go
```diff
@@ -18,11 +23,198 @@ const (
...
+	}
+	if err = bson.Unmarshal(req.Params.Options, opts); err != nil {
```
Kimchelly, line 50: See comment about scriptingCreateRequest.

remote/mdb_service_scripting.go
```diff
@@ -18,11 +23,198 @@ const (
...
+
+func (s *mdbService) scriptingSetup(ctx context.Context, w io.Writer, msg mongowire.Message) {
+	req := &scriptingSetupRequest{}
+	if !s.serviceScriptingRequest(ctx, w, msg, req, ScriptingSetupCommand) {
+		return
+	}
+
+	harness := s.getHarness(ctx, w, req.ID, ScriptingSetupCommand)
+	if harness == nil {
+		return
```
Kimchelly, line 71: I see you checked for nil harness cache a bunch of times. Given that the only entry point to make an `mdbService` is through `StartMDBService()`, is it really necessary to do lots of nil checking for whether the service has a harness when we know it will have one?

remote/mdb_io.go
julianedwards: You can't bson encode an interface but I can bson encode type `interface{}`. The problem then is that I would need to do a type conversion to the underlying type on the server side to get it back to the `options.ScriptingHarness`.

remote/mdb_service_scripting.go
julianedwards: Maybe not? But this also checks for the ID in that harness, which could not exist.

Kimchelly: Fair. I think it's fine to handle that the ID might not exist but the cache itself should definitely exist. Can we get rid of just the checks for whether `s.harnessCache` is `nil` since we already know it's non-nil?

remote/mdb_io.go
Kimchelly: I think this is fair but it might be more cumbersome for end users to have to serialize their inputs to raw BSON bytes before being able to send a scripting request. All of the requests were written in such a way that you could look at the struct definition and determine what the arguments were without needing to do some data transformation on the input. If Options was an `interface{}` type it would accept anything as long as it could be type casted into one of the scripting option types since they already have to specify the type anyways.

remote/client_test.go
```diff
@@ -1125,3 +1253,11 @@ func TestManager(t *testing.T) {
 		})
 	}
 }
+
+func createScriptingHarness(ctx context.Context, t *testing.T, client Manager, dir string) scripting.Harness {
```
tychoish, line 1257: this should have a different name.

remote/mdb_io.go
```diff
@@ -283,3 +284,88 @@ type loggingCacheLenRequest struct {
...
+type scriptingBuildRequest struct {
+	Params struct {
+		ID   string   `bson:"id"`
+		Dir  string   `bson:"dir"`
+		Args []string `bson:"args"`
+	} `bson:"build_scripting"`
+}
+
+type scriptingBuildResponse struct {
+	shell.ErrorResponse `bson:"error_response,inline"`
```
tychoish, line 342: are we using bson parsers that respect inline?

remote/mdb_service_scripting.go
```diff
@@ -5,6 +5,11 @@ import (
 	"io"
 
 	"github.com/evergreen-ci/mrpc/mongowire"
+	"github.com/evergreen-ci/mrpc/shell"
+	"github.com/mongodb/jasper/options"
+	"github.com/mongodb/jasper/scripting"
+	"github.com/pkg/errors"
+	"gopkg.in/mgo.v2/bson"
```
tychoish, line 12: we shouldn't introduce new users of the legacy library. 

I'd vaguely prefer if things used birch, but if we need reflection (and/or existing parts of the service uses reflection), then we can/should use the new library

remote/mdb_service_scripting.go
tychoish: does the rest of the mrdb service use the bson library?

remote/mdb_io.go
julianedwards: Yes, inline is respected here.

remote/mdb_service_scripting.go
Kimchelly: Yes. It's because birch doesn't have an equivalent `bson.Marshal` function and there's no easy way to make a birch document from an arbitrary struct without first serializing it to BSON in the first place.

julianedwards:
@Kimchelly are you good?

PR #176, opened 2020-04-23
MAKE-1179: reduce number of usages of yes command

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1179

I still have difficulty reproducing this since it occurs maybe 10% of the time and only in the agent. My best hunch right now is that using the `yes` command on Windows is expensive. I think I've seen this before, when I was originally writing the benchmarks.

The cause doesn't seem to be in `TestGetInMemoryLogStream` - it just happens to run after several tests running `yes` finish, which seems to slow down runtime. Since basic process does not clean up the transition goroutine until the process actually exits, other tests may be leaving too many `yes` commands running after said tests are done (they do eventually clean up, but not until they hit their timeout value). Since these tests really just want to simulate any long-running command (so they don't necessarily need `yes`), I converted these to `sleep` commands, which are seemingly not as taxing on the system.

* Remove usages of `yes` in tests.
* Do some better error wrapping in closers.
* Add some omitempties for optional values in structs.
* Remove windows-64-vs2010 distros.

### DIFF
--- a/evergreen.yaml
+++ b/evergreen.yaml
@@ -197,11 +197,9 @@ buildvariants:
     display_name: Windows
     run_on:
       - windows-64-vs2015-small
-      - windows-64-vs2015-test
-      - windows-64-vs2015-compile
       - windows-64-vs2015-large
-      - windows-64-vs2013-compile
-      - windows-64-vs2013-test
+      - windows-64-vs2017-small
+      - windows-64-vs2017-large
     expansions:
       DISABLE_COVERAGE: true
       GOROOT: C:/golang/go1.11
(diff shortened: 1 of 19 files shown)


evergreen.yaml
```diff
@@ -200,8 +200,6 @@ buildvariants:
       - windows-64-vs2015-test
       - windows-64-vs2015-compile
       - windows-64-vs2015-large
-      - windows-64-vs2010-compile
-      - windows-64-vs2010-test
```
tychoish, line 204: can you take out 2013 and add 2017 here too?

PR #177, opened 2020-04-28
MAKE-1282: fix RPC handling for time conversions and add start and end to process info

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1282

* Add StartAt and EndAt for process info from RPC.
* Error instead of panic when converting timestamps to/from protobuf RPC types.
* Fix error codes in RPC scripting interface
* Documentation for RPC scripting


evergreen.yaml
```diff
@@ -197,13 +197,9 @@ buildvariants:
...
       - windows-64-vs2015-small
-      - windows-64-vs2015-test
-      - windows-64-vs2015-compile
       - windows-64-vs2015-large
-      - windows-64-vs2010-compile
-      - windows-64-vs2010-test
-      - windows-64-vs2013-compile
-      - windows-64-vs2013-test
+      - windows-64-vs2017-small
+      - windows-64-vs2017-large
```
tychoish, line 202: I'd put 2017 before 2015

PR #178, opened 2020-05-04
MAKE-1226: implement processes within Docker containers

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1226

Currently tested on ubuntu. I would prefer to also run with the race detector, but it seems like archlinux-small doesn't have Docker installed. Docker is tested against the process/manager interfaces, but I'll write tests specifically for the Docker executor in [MAKE-1300](https://jira.mongodb.org/browse/MAKE-1300).

* Vendor github.com/docker/docker and github.com/evergreen-ci/utility.
* Containerized processes can be constructed by specifying Docker options in `options.Create`.
* Add Docker manager that just specifies default Docker options for process creation.
* Some code refactoring/simplification/documentation for Executor behavior.
* Unrelated fix to remove unnecessary duplicate Remote options in Command.


options/docker.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Docker encapsulates options related to connecting to a Docker daemon.
+type Docker struct {
+	Host       string
+	Port       int
+	APIVersion string
+	Image      string
+	// Platform refers to the major operating system on which the Docker
+	// container runs.
+	Platform string
```
tychoish, line 22: should we expect to add other docker runtime options? or at least provide provisions along those lines.

options/docker.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Copy returns a copy of the options for only the exported fields.
+func (opts *Docker) Copy() *Docker {
+	optsCopy := *opts
+	return &optsCopy
+}
+
+// DockerPlatforms returns all supported platforms that can run Docker
+// processes.
+func DockerPlatforms() []string {
```
tychoish, line 51: I'd be slightly more inclined to have this be PlatformSupportsDocker(string) bool

util/slice.go
```diff
@@ -0,0 +1,16 @@
+package util
+
+// StringSliceContains determines if a string is in a slice
+func StringSliceContains(slice []string, item string) bool {
```
tychoish, line 4: use evergreen-ci/utility

internal/executor/executor.go
```diff
@@ -6,24 +6,108 @@ import (
...
+	return lower < s && s < upper
+}
+
+func (s Status) BetweenInclusive(lower Status, upper Status) bool {
+	return lower <= s && s <= upper
+}
+
+const (
+	// Unknown means the process
+	Unknown = iota
```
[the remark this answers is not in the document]
Kimchelly: The`iota` is actually necessary in this case because I defined a `Status` type enum instead of an `int`, and Go isn't smart enough to be able to assign to non-`int` types unless you use `iota` every single time. Regardless, assigning `iota` on every constant correctly assigns incremental numbers to each constant.

options/docker.go
Kimchelly: Yeah, I didn't add any options except the bare minimum necessary to verify that Docker containers work. We can add more when we need them.

evergreen.yaml
```diff
@@ -36,6 +36,25 @@ functions:
...
+    params:
+      binary: docker
+      args: ["pull", "ubuntu"]
+  cleanup-docker-containers:
+    command: shell.exec
+    type: setup
+    params:
+      shell: bash
+      script: |
+        docker rm -f $(docker ps -a -q)
```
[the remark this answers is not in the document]
Kimchelly: I moved the Docker-related things to the makefile.

util/slice.go
Kimchelly: Done.

options/docker.go
Kimchelly: Done.

PR #179, opened 2020-05-05
MAKE-1152: Support dependency injection of logging backends in jasper 

julianedwards:


options/logger_registry.go
```diff
@@ -0,0 +1,112 @@
...
+
+		return nil
+	default:
+		return errors.Errorf("unsupported format '%s'", f)
+	}
+}
+
+// LoggerProducer produces a Logger interface backed by a grip logger.
+type LoggerProducer interface {
+	Configure() (send.Sender, error)
```
tychoish, line 108: you added a validate method here, and I like that, I think it should be part of the interface. 

I also think that you have to specify the type here here, so that you can write an implementation in the marshalers that do the right thing. (see my Set() method.)

options/logger_registry.go
```diff
@@ -0,0 +1,112 @@
...
+	case LogConfigFormatBSON, LogConfigFormatJSON:
+	default:
+		return errors.New("unknown log config format")
+	}
+
+	return nil
+}
+
+// Unmarshal unmarshals the data into out based on the log config format.
+func (f LogConfigFormat) Unmarshal(data []byte, out interface{}) error {
```
tychoish, line 85: I think this should become private. 

also I think out can be of LoggerProducer type

options/logger_registry.go
```diff
@@ -0,0 +1,112 @@
...
+type LogConfigFormat string
+
+const (
+	LogConfigFormatBSON LogConfigFormat = "BSON"
+	LogConfigFormatJSON LogConfigFormat = "JSON"
+)
+
+func (f LogConfigFormat) Validate() error {
+	switch f {
+	case LogConfigFormatBSON, LogConfigFormatJSON:
```
tychoish, line 76: I'd be explicit about returning nil here.

options/logger_registry.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+}
+
+// LoggerProducer produces a Logger interface backed by a grip logger.
+type LoggerProducer interface {
+	Configure() (send.Sender, error)
+}
+
+// LoggerFactory creates a new instance of a LoggerProducer implementation.
+type LoggerFactory func() LoggerProducer
```
tychoish, line 112: I'd call this LoggerProducerFactory

options/logger.go
```diff
@@ -82,190 +55,154 @@ func (f LogFormat) MakeFormatter() (send.MessageFormatter, error) {
...
 }
 
-// Validate ensures that BufferOptions is valid.
-func (opts Buffer) Validate() error {
-	if opts.Buffered && opts.Duration < 0 || opts.MaxSize < 0 {
-		return errors.New("cannot have negative buffer duration or size")
+func (lc *LoggerConfig) Validate() error {
+	catcher := grip.NewBasicCatcher()
+
+	// TODO: maybe check for nil or empty Data
```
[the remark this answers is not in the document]
julianedwards: but what if someone wants to implement a logger that doesn't need any config options?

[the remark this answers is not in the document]
tychoish: then the empty but non-nil is useful? I could go either way.

options/logger_registry.go
julianedwards: I think the validate method should be build in to the Configure() since it already returns an error (and its up to the implementation to decided what needs to be validated, if anything at all). I changed the LoggerRegistry to use the LoggerProducer.Type() method as the name (so we don't need to check if after the fact).

options/logger.go
```diff
@@ -82,190 +55,154 @@ func (f LogFormat) MakeFormatter() (send.MessageFormatter, error) {
...
-// buffered and the duration and size of the respective buffer in the case that
-// it should be.
-type Buffer struct {
-	Buffered bool          `bson:"buffered" json:"buffered" yaml:"buffered"`
-	Duration time.Duration `bson:"duration" json:"duration" yaml:"duration"`
-	MaxSize  int           `bson:"max_size" json:"max_size" yaml:"max_size"`
+type LoggerConfig struct {
+	Type     string          `json:"type" bson:"type"`
+	Format   LogConfigFormat `json:"format" bson:"format"`
+	Data     []byte          `json:"data" bson:"data"`
```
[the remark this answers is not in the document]
julianedwards: This is no longer in the output so I don't think its necessary

options/logger.go
```diff
@@ -82,190 +56,171 @@ func (f LogFormat) MakeFormatter() (send.MessageFormatter, error) {
...
-		return errors.New("cannot have negative buffer duration or size")
+// Validate ensures the LoggerConfig is valid.
+func (lc *LoggerConfig) Validate() error {
+	catcher := grip.NewBasicCatcher()
+
+	catcher.NewWhen(lc.Type == "", "cannot have empty logger type")
+	catcher.Add(lc.Format.Validate())
+
+	if lc.Registry == nil {
+		lc.Registry = globalLoggerRegistry
```
Kimchelly, line 88: It seems kind of weird that every single LoggerConfig instance contains the global logging registry. Is there reason we can't just get rid of this field and just get logger factories directly from the global registry?

options/logger.go
```diff
@@ -82,190 +56,171 @@ func (f LogFormat) MakeFormatter() (send.MessageFormatter, error) {
...
+// LoggerConfig represents the necessary information to construct a new grip
+// send.Sender. LoggerConfig implements the bson.MarshalBSON and
+// json.MarshalJSON interfaces.
+type LoggerConfig struct {
+	Type     string          `json:"type" bson:"type"`
+	Format   LogConfigFormat `json:"format" bson:"format"`
+	Config   LogConfig       `json:"config" bson:"config"`
+	Registry LoggerRegistry  `json:"-" bson:"-"`
+
+	producer LoggerProducer
```
Kimchelly, line 76: Can we get rid of this since it's not being set anywhere and all we really need is the sender it produces?

options/logger_producers.go
```diff
@@ -0,0 +1,319 @@
...
+		LogInMemory:      NewInMemoryLoggerProducer,
+		LogSplunk:        NewSplunkLoggerProducer,
+		LogBuildloggerV2: NewBuildloggerV2LoggerProducer,
+	},
+}
+
+// GetGlobalLoggerRegistry returns the global logger registry.
+func GetGlobalLoggerRegistry() LoggerRegistry { return globalLoggerRegistry }
+
+///////////////////////////////////////////////////////////////////////////////
```
Kimchelly, line 24: Can we just split the different implementations into different files instead of use doc comments to separate implementations?

options/logger.go
```diff
@@ -1,49 +1,23 @@
...
 const (
 	// DefaultLogName is the default name for logs emitted by Jasper.
 	DefaultLogName = "jasper"
 )
 
-// LogFormat specifies a certain format for logging by Jasper.
-// See the documentation for grip/send for more information on the various
-// LogFormat's.
+// LogFormat specifies a certain format for logging by Jasper. See the
+// documentation for grip/send for more information on the various LogFormat's.
```
Kimchelly, line 20: LogFormats*

options/logger.go
```diff
@@ -82,190 +56,171 @@ func (f LogFormat) MakeFormatter() (send.MessageFormatter, error) {
...
+	return catcher.Resolve()
 }
 
-// Log contains options related to the logging done by Jasper.
-//
-// By default, logger reads from both standard output and standard error.
-type Log struct {
-	BufferOptions        Buffer                    `json:"buffer_options,omitempty" bson:"buffer_options"`
-	BuildloggerOptions   send.BuildloggerConfig    `json:"buildlogger_options,omitempty" bson:"buildlogger_options"`
-	BuildloggerV3Options timber.LoggerOptions      `json:"buildlogger_v3_options" bson:"buildlogger_v_3_options"`
```
Kimchelly, line 108: Since the buildlogger v3 isn't implemented anymore with timber, is the expectation that we're removing the vendored timber so external packages (i.e. evergreen) implement it?

If so, can we remove the timber references from `make vendor-clean` and remove the vendor directory?

options/logger_producers.go
```diff
@@ -0,0 +1,319 @@
+package options
+
+import (
+	"github.com/mongodb/grip"
+	"github.com/mongodb/grip/send"
+	"github.com/pkg/errors"
+)
+
+var globalLoggerRegistry LoggerRegistry = &basicLoggerRegistry{
```
tychoish, line 9: can we put these function in the registry file?

options/output.go
```diff
@@ -238,11 +234,6 @@ func (o *Output) Close() error {
 	if o.errorSender != nil && (o.SuppressOutput || o.SendOutputToError) {
 		catcher.Wrap(o.errorSender.Sender.Close(), "problem closing wrapped error sender")
 	}
-	// Since loggers are owned by this process, we close the loggers to clean up
```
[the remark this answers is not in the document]
tychoish: +1

options/logger.go
tychoish: yes!

options/logger.go
```diff
@@ -78,194 +52,264 @@ func (f LogFormat) MakeFormatter() (send.MessageFormatter, error) {
...
-	Level                send.LevelInfo            `json:"level,omitempty" bson:"level"`
-	SplunkOptions        send.SplunkConnectionInfo `json:"splunk_options,omitempty" bson:"splunk_options"`
-	SumoEndpoint         string                    `json:"sumo_endpoint,omitempty" bson:"sumo_endpoint"`
+// LoggerConfig represents the necessary information to construct a new grip
+// send.Sender. LoggerConfig implements the bson.MarshalBSON and
+// json.MarshalJSON interfaces.
+type LoggerConfig struct {
+	Type     string                `json:"type" bson:"type"`
+	Format   RawLoggerConfigFormat `json:"format" bson:"format"`
+	Config   []byte                `json:"config" bson:"config"`
```
tychoish, line 107: I think that this needs to be a type alias with marshal functions so that when we marshal the logger config, this doesn't end up coming out as numbers. Putting the on the LoggerConfig object means that you have to duplicate the structure and tags in an autonomous struct, and if you put it here, and produce the byte slice in Set() then it's possible to modify LoggerConfig in the future (if we want/need to) without needing to modify the marshaling functions.

options/logger.go
```diff
@@ -78,194 +52,264 @@ func (f LogFormat) MakeFormatter() (send.MessageFormatter, error) {
...
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrap(err, "problem producing logger config")
 		}
-	default:
-		return nil, errors.New("unknown log type")
+		lc.Config = data
+	}
+
+	return json.Marshal(struct {
```
tychoish, line 220: I think duplicating this structure is a bit fraught, and I'm slightly worried that the Config fields will still hold an array of numbers, see earlier comment.

options/logger.go
```diff
@@ -77,95 +77,120 @@ func (f RawLoggerConfigFormat) Validate() error {
...
+
+func (lc *RawLoggerConfig) MarshalJSON() ([]byte, error) {
+	if len(*lc) > 0 {
+		return *lc, nil
+	}
+
+	return json.Marshal([]byte(*lc))
+}
+
+func (lc *RawLoggerConfig) UnmarshalJSON(b []byte) error {
```
tychoish, line 113: do you need to have unmarshal functions? I think it should just work...

options/logger.go
```diff
@@ -77,95 +77,120 @@ func (f RawLoggerConfigFormat) Validate() error {
...
+
+func (lc RawLoggerConfig) MarshalBSON() ([]byte, error) {
+	if len(lc) > 0 {
+		return lc, nil
+	}
+
+	return bson.Marshal([]byte(lc))
+}
+
+func (lc *RawLoggerConfig) UnmarshalBSON(b []byte) error {
```
tychoish, line 126: I think you can omit this?

options/logger.go
```diff
@@ -77,95 +77,120 @@ func (f RawLoggerConfigFormat) Validate() error {
...
+// RawLoggerConfig wraps []byte and implements the json and bson Marshaler and
+// Unmarshaler interfaces.
+type RawLoggerConfig []byte
+
+func (lc *RawLoggerConfig) MarshalJSON() ([]byte, error) {
+	if len(*lc) > 0 {
+		return *lc, nil
+	}
+
+	return json.Marshal([]byte(*lc))
```
tychoish, line 110: I suspect you're never hitting this...

options/logger.go
```diff
@@ -77,95 +77,120 @@ func (f RawLoggerConfigFormat) Validate() error {
...
+// Producer returns the underlying logger producer for this logger config,
 // which may be nil.
-func (lc *LoggerConfig) GetProducer() LoggerProducer { return lc.producer }
+func (lc *LoggerConfig) Producer() LoggerProducer { return lc.producer }
+
+// Type returns the type string.
+func (lc *LoggerConfig) Type() string { return lc.info.Type }
+
+// Config returns the raw logger config.
+func (lc *LoggerConfig) Config() RawLoggerConfig { return lc.info.Config }
```
tychoish, line 186: is there anyway the interface can return the result of "resolve" rather than a byte slice

options/logger.go
julianedwards: no it fails, i originally didn't have them and it complains that it can't unmarshal

options/logger.go
julianedwards: actually we don't need this method, removing it

PR #180, opened 2020-05-11
MAKE-1280: generate evergreen config for golang

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1280
https://jira.mongodb.org/browse/MAKE-1319
https://jira.mongodb.org/browse/MAKE-1302

# Changes
* Basic YAML models for configuring go tests.
* Auto-discover Go packages to test.
* Generate basic shrub config.
    * For each package referenced in each variant, we have to make one task because options can be specified at the variant or package level and we have no way of applying different command-level options to different variants without making unique tasks for each package in a variant.
    * Only create a task group if the variant runs at least 4 tasks (i.e. 4 golang packages)
* Add bare-bones CLI that basically just calls the generator and outputs the JSON evergreen config.

# Sample input file
```
---
env:
  GOPATH: gopath
  GOROOT: /opt/golang/go1.13

packages:
  - name: jasper
    path: .
    tags: ["lint", "test"]
    options: ["-short"]

  - path: model
    tags: ["lint", "test"]

  - path: model/internal
    tags: ["test"]

  - name: util
    path: util
    tags: ["lint"]

  - name: util-for-testing
    path: testutil
    tags: ["lint"]

variants:
  - name: ubuntu
    distros:
      - ubuntu1604-small
      - ubuntu1604-large
    packages:
      - tag: test

  - name: race-detector
    distros:
      - archlinux-small
      - archlinux-large
    packages: 
      - tag: test
      - name: util
      - path: service # implicitly defined after discovering packages
    options: ["-race", "-short"]

  - name: lint
    distros:
      - archlinux-small
      - archlinux-large
    packages: 
      - tag: lint
      - path: model/internal
    options: ["-cover"]
```

# TODO
* ~Write more thorough tests for shrub config generation.~ Will be done in [MAKE-1319](https://jira.mongodb.org/browse/MAKE-1320)
* ~Handle merging identical test options to `go test` better.~ Done.
* ~Maybe more options for the jasper generator file.~ Will be done once we have a better idea of what the complete generator looks like.


tychoish:
things that also should happen: 
- making a parsing model or file organization that allows putting things in multiple files.
- making a global setup area that includes environment but also things like packages that you're going to add to to the scripting environment (in addition to environment variables.) 
- splitting out the mapping between dimension (variants) and distro so that happens in a different file/area.
- we should have an answer to what the best way to "tag a, but not b" 
- fields should have default resolutions, like package name/field shouldn't need to be specified if they're both the same. path could/should default to CWD, etc.

buildsystem/generator/golang.go
```diff
@@ -0,0 +1,597 @@
...
+type GolangVariant struct {
+	Name    string   `yaml:"name"`
+	Distros []string `yaml:"distros"`
+	// Packages lists the names or paths of the package relative to the root
+	// package, which can be one of the following:
+	// - A name of a package explicitly listed in the packages section.
+	// - A path of a package explicitly listed in the packages section.
+	// - A path of a package that matches a path to a directory containing tests
+	//   relative to the root package.
+	// - A package tag, beginning with a period (e.g. ".tag").
```
tychoish, line 514: are we doing .tags still?

buildsystem/generator/golang.go
```diff
@@ -0,0 +1,597 @@
...
+			pkgTags[tag] = struct{}{}
+		}
+	}
+	if gv.Options != nil {
+		catcher.Wrap(gv.Options.Validate(), "invalid runtime options")
+	}
+	return catcher.Resolve()
+}
+
+type VariantPackage struct {
```
tychoish, line 557: I'd put all the tags on this just be be safe.

buildsystem/generator/golang.go
```diff
@@ -0,0 +1,597 @@
...
+// GolangRuntimeOptions specify additional options to modify behavior of runtime
+// execution.
+type GolangRuntimeOptions struct {
+	// TODO: add more go test flags? It might be somewhat onerous to maintain
+	// all the possible go test flags here, but parsing the individual flags to
+	// merge them is also painful.
+	Args []string `yaml:"args,omitempty"`
+}
+
+// TODO: validate options duplication.
```
tychoish, line 585: yes this would be cool

options/scripting_golang.go
```diff
@@ -9,16 +9,22 @@ import (
...
 type ScriptingGolang struct {
-	Gopath         string            `bson:"gopath" json:"gopath" yaml:"gopath"`
-	Goroot         string            `bson:"goroot" json:"goroot" yaml:"goroot"`
-	Packages       []string          `bson:"packages" json:"packages" yaml:"packages"`
-	Context        string            `bson:"context" json:"context" yaml:"context"`
-	WithUpdate     bool              `bson:"with_update" json:"with_update" yaml:"with_update"`
+	Gopath string `bson:"gopath" json:"gopath" yaml:"gopath"`
+	Goroot string `bson:"goroot" json:"goroot" yaml:"goroot"`
+	// Packages describes the required packages for running scripts.
+	// TODO: better support for go modules for getting specific versions?
```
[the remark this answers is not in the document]
Kimchelly: I'm pretty sure you have to run `go mod download` to get the dependencies, but I wouldn't know for sure since we never use modules.

options/scripting_python.go
```diff
@@ -15,32 +15,43 @@ import (
...
+	// RequirementsPath specifies a path to the requirements.txt containing the
+	// required packages.
+	RequirementsPath string `bson:"requirements_path" json:"requirements_path" yaml:"requirements_path"`
+	// Packages specifies all required packages for running scripts.
+	Packages []string `bson:"packages" json:"packages" yaml:"packages"`
+	// AddTestRequirements determines whether or not to install the dependencies
+	// necessary for executing scripts if they are not already installed.
+	AddTestRequirements bool `bson:"add_test_reqs" json:"add_test_reqs" yaml:"add_test_reqs"`
+	// TODO: should there be an equivalent UpdatePackages flag like for golang
+	// (i.e. pip install --upgrade)?
```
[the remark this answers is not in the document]
Kimchelly: Done.

options/scripting_python.go
```diff
@@ -15,32 +15,43 @@ import (
...
+	// TODO: should this be an enum to handle when it's not just python 2 vs 3?
```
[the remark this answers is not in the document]
Kimchelly: I'm going to leave this todo for now since it's an open question.

buildsystem/generator/golang.go
Kimchelly: No. I removed this comment.

buildsystem/generator/golang.go
Kimchelly: The tags must go on the package itself. This is just used in the variant definition to refer to a package definition by name/path/tag.

buildsystem/generator/golang.go
Kimchelly: Cool, but would be ugly and I don't think it's worth it to deal with for now since it would mostly devolve into string parsing/checking/deduplication and isn't critical for the config generation right now.

PR #183, opened 2020-05-20
MAKE-1152: Buildloggerv3 rpc

julianedwards:


options/logger.go
```diff
@@ -132,6 +133,19 @@ type loggerConfigInfo struct {
 	Config RawLoggerConfig       `json:"config" bson:"config"`
 }
 
+// NewLoggerConfig returns a LoggerConfig with the given info, this function
+// expects a raw logger config as a byte slice. When using a LoggerProducer
+// directly, use LoggerConfig's Set function.
+func NewLoggerConfig(producerType string, format RawLoggerConfigFormat, config []byte) *LoggerConfig {
```
tychoish, line 139: can you pass a LoggerProducer rather than it as JSON?

jasper.proto
```diff
@@ -94,6 +95,36 @@ message BuildloggerV2Options {
...
+  bool disable_new_line_check = 16;
+  string rpc_address = 17;
+  bool insecure = 18;
+  string ca_file = 19;
+  string cert_file = 20;
+  string key_file = 21;
+}
+
+message BuildloggerV3Options {
+  BuildloggerV3Info buildloggerv3 = 1;
```
tychoish, line 123: why split between info and options? 

maybe split the first 9 fields out?

options/logger.go
julianedwards: We have Set for that. This is a special case when you have the raw log config but potentially not the logger producer (i.e. buildlogger v3's setup being in timber)

jasper.proto
julianedwards: Don't have much flexibility here... this has to copy exactly whats [here](https://github.com/evergreen-ci/timber/blob/master/logger_producer.go#L11-L15) in timber so it can marshal correctly.

PR #184, opened 2020-05-20
MAKE-1315: add generator for Make and support multiple files (WIP)

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1315

## 
* Write basic skeleton code for Make generator and evergreen config generation.
* Vendor `github.com/yargevad/filepathx` for globbing support with `**`.
* Add CLI support for generating Make from single file or from control file that can reference definition files for:
    * Variant-distro mappings
    * Variant-task mappings
    * Task definitions
    * Environment variables

## Sample Make Build File
```yaml
---
environment:
  DISABLE_COVERAGE: 1

tasks:
  test-util:
    tags: ["lint", "test", "race"]
    targets:
      - setup
      - test-util
  test-model:
    tags: ["lint", "test", "race"]
    targets:
      - setup-mongodb
      - test-model
  test-mock:
    tags: ["lint"]
    targets:
      - test-mock

variants:
  ubuntu1604:
    distros: ["ubuntu1604-small", "ubuntu1604-large"]
    tasks:
      - tag: test
      - name: test-mock
  race_detector:
    distros: ["archlinux-small", "archlinux-large"]
    environment:
      RACE_DETECTOR: true
    tasks:
      - tag: race
  lint:
    distros: ["archlinux-small", "archlinux-large"]
    environment:
      DISABLE_COVERAGE: ""
    tasks:
      - tag: lint
```

## TODO
* Allowing Make targets to be configured (e.g. tag groups of targets).
* Allowing arbitrary flags to be passed to Make.
* I'm going to do a minor refactor of the Go generator to use maps like the Make generator does to reduce the amount of duplicate-definition checking and simplify a lot of the code.
* Tests


buildsystem/generator/control_util.go
```diff
@@ -0,0 +1,23 @@
+package generator
+
+import (
+	"github.com/pkg/errors"
+	"github.com/yargevad/filepathx"
+)
+
+func withGlobMatches(patterns []string, op func(file string) error) error {
```
tychoish, line 8: evergreen uses this git-ignore parsing module module, and I think we should standardize on that

buildsystem/generator/make.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/evergreen-ci/utility"
+	"github.com/mongodb/grip"
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
+)
+
+// Make represents an evergreen config generator for Makefile-based projects.
+type Make struct {
+	Tasks    map[string]MakeTask    `yaml:"tasks,omitempty"`
+	Variants map[string]MakeVariant `yaml:"variants"`
```
tychoish, line 16: should these be lists instead?

buildsystem/generator/make.go
```diff
@@ -0,0 +1,300 @@
...
+	// Environment defines global environment variables. Definitions can be
+	// overridden at the task or variant level.
+	Environment      map[string]string `yaml:"environment,omitempty"`
+	WorkingDirectory string            `yaml:"-"`
+}
+
+// NewMake creates a new evergreen config generator for Make from a single file
+// that contains all the necessary generation information.
+func NewMake(file string, workingDir string) (*Make, error) {
```
tychoish, line 25: is it time to bring back the interface. I'm iffy about having the *Make type exported.

buildsystem/generator/make.go
tychoish: I'm also, not sure about this, but it seems like it might make sense to split out generator implementation/models (as with options), but I'm not sure.

buildsystem/generator/make.go
```diff
@@ -0,0 +1,300 @@
+package generator
+
+import (
+	"io/ioutil"
+
+	"github.com/evergreen-ci/shrub"
+	"github.com/evergreen-ci/utility"
+	"github.com/mongodb/grip"
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
```
tychoish, line 10: I suspect you can just use ReadYAML/WriteYAML from utility.

buildsystem/generator/make.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+		tasks := map[string]MakeTask{mvt.Name: *task}
+		return &taskReferenceResult{tasks: tasks, ref: mvt.Name}, nil
+	}
+
+	return nil, errors.New("empty package reference")
+}
+
+func (m *Make) Generate() (*shrub.Configuration, error) {
+	conf, err := shrub.BuildConfiguration(func(c *shrub.Configuration) {
```
tychoish, line 152: I might just return this directly.

buildsystem/generator/make.go
```diff
@@ -0,0 +1,300 @@
...
+		for vName, mv := range m.Variants {
+			variant := c.Variant(vName)
+			variant.DistroRunOn = mv.Distros
+
+			var tasksForVariant []*shrub.Task
+			for _, mvt := range mv.Tasks {
+				res, err := m.getTasksAndRef(mvt)
+				if err != nil {
+					panic(err)
```
tychoish, line 161: just to check, but the BuildConfiguration handles panics?

buildsystem/generator/make.go
```diff
@@ -0,0 +1,300 @@
...
+	tasks := map[string]MakeTask{}
+	for name, mt := range m.Tasks {
+		if utility.StringSliceContains(mt.Tags, tag) {
+			tasks[name] = mt
+		}
+	}
+	return tasks
+}
+
+type MakeTask struct {
```
tychoish, line 234: docstring

glide.yaml
```diff
@@ -17,7 +17,7 @@ import:
   - package: github.com/tychoish/bond
   - package: github.com/tychoish/lru
   - package: github.com/mholt/archiver
-  - package: github.com/montanaflynn/stats
```
tychoish, line 20: should we delete?

buildsystem/generator/control_util.go
Kimchelly: I looked at [the package](https://godoc.org/github.com/sabhiram/go-gitignore) and the package interface seems to only be good at answering whether or not to ignore a particular file and is not actually able to find files. You would still have to roll your own `**` functionality to search for files.

buildsystem/generator/make.go
Kimchelly: I would prefer to `yaml.UnmarshalStrict` rather than just `yaml.Unmarshal`, which is what utility does.

buildsystem/generator/make.go
Kimchelly: It would be significantly more difficult for them to be slices since you would have to do lots of duplicate checking when merging file parts and wouldn't provide a great deal of use to make them slices since we're assuming there's 1 definition for each task/variant. I don't think the YAML file reads particularly differently and to me is actually prefereable if you look at the example.

We could split the generator models, but they would look identical so it would just be copy-paste.

buildsystem/generator/make.go
Kimchelly: We could interface this but the only commonality would be the `Generate` function. There's almost no shared functionality between building an evergreen config from a Make generator and building an evergreen config from a Golang generator except their output. The actual implementation needs of the different generators have almost no similarity because they need different definition files to be merged in.

I would prefer to just duck-type it and if someone wants to say that the only functionality it has is `Generate()`, then they can define their own interface that gates the Make/Golang generator functionality. I don't actually like to return interfaces when you could just receive the concrete type and compile-time check that it satisfies whatever interface you want to put it in.

Also, returning a limited interface would make the difficult to use unless we made a separate model for the YAML file representation (with no interface), marshalled that representation into raw bytes, and then passed those raw bytes into the Make generator to generate evergreen config.

buildsystem/generator/make.go
Kimchelly: I want to wrap the error.

buildsystem/generator/make.go
Kimchelly: Yeah, it's the only way it can produce errors.

buildsystem/generator/make.go
Kimchelly: Done.

glide.yaml
Kimchelly: It's unused, so I deleted it.

buildsystem/model/golang_control.go
```diff
@@ -0,0 +1,120 @@
+package model
+
+import (
+	"github.com/evergreen-ci/utility"
+	"github.com/mongodb/grip"
+	"github.com/pkg/errors"
+)
+
+type GolangControl struct {
+	// TODO: this configuration is required to know the full go package name but
```
julianedwards, line 10: can we make a ticket for this and put it in the TODO comment?

buildsystem/model/common.go
```diff
@@ -0,0 +1,31 @@
...
+	catcher.NewWhen(vd.Name == "", "missing variant name")
+	catcher.NewWhen(len(vd.Distros) == 0, "need to specify at least one distro")
+	return catcher.Resolve()
+}
+
+// MergeEnvironments returns the merged environment variable mappings where the
+// given input environments are ordered in increasing priority. If there are
+// duplicate environment variables, in the environments, the variable definition
+// of the higher priority environment takes precedence.
+func MergeEnvironments(envsByPriority ...map[string]string) map[string]string {
```
julianedwards, line 23: So this assumes that the envs, or `map[string]string` objects, in envByPriority are in increasing order by priority?

buildsystem/model/golang_control.go
```diff
@@ -0,0 +1,120 @@
+package model
```
julianedwards, line 1: add documentation to exported types and funcs

Kimchelly: Done.

PR #189, opened 2020-06-08
MAKE-1341: fix jasper benchmarks 

julianedwards:


makefile
```diff
@@ -56,7 +56,7 @@ $(buildDir)/run-linter:cmd/run-linter/run-linter.go $(buildDir)/.lintSetup $(bui
 
 # benchmark setup targets
 $(buildDir)/run-benchmarks:cmd/run-benchmarks/run_benchmarks.go $(buildDir)
-	$(goEnv) $(gobin) build -o $@ $<
+	$(goEnv) $(gobin) build -ldflags "-w" -o $@ $<
```
Kimchelly, line 59: Why the ldflags?

julianedwards: it doesn't run on catalina without them

Kimchelly: What do you mean "doesn't run"? I just tried running it without ldflags and it started fine.

julianedwards: `dyld: malformed mach-o image: segment __DWARF has vmsize < filesize`

julianedwards: i get this error without the ldflags

Kimchelly: I can't reproduce. Can you remove it and we can look at this later?

PR #193, opened 2020-06-17
MAKE-1371: allow package discovery to include source-only directories

Kimchelly:
JIRA: https://jira.mongodb.org/browse/MAKE-1371

The metabuild has a feature for go projects to automatically create task definitions on a per-package basis without you having to explicitly list all the packages (it does this by walking the project directory tree). Right now, the generator will only auto-discover go packages if they have `*_test.go` files in them. This adds an option to auto-discover any package that contains a go file in it (`*.go`).

### DIFF
--- a/metabuild/generator/golang.go
+++ b/metabuild/generator/golang.go
@@ -45,12 +45,8 @@ func (g *Golang) Generate() (*shrub.Configuration, error) {
 				tasksForVariant = append(tasksForVariant, newTasks...)
 			}
 
-			gopath := g.Environment["GOPATH"]
-			// kim: TODO: test that this is modified if the variant specifies a
-			// GOPATH.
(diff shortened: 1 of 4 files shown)


metabuild/generator/golang.go
```diff
@@ -45,10 +45,8 @@ func (g *Golang) Generate() (*shrub.Configuration, error) {
 				tasksForVariant = append(tasksForVariant, newTasks...)
 			}
 
-			gopath := g.Environment["GOPATH"]
-			if val := gv.Environment["GOPATH"]; val != "" {
-				gopath = val
-			}
+			env := model.MergeEnvironments(g.Environment, gv.Environment)
```
julianedwards, line 48: nice!

PR #195, opened 2020-07-08
EVG-9706 use regexp

ybrill:
The parsing was relying on a space after the pid, which wasn't true on rhel

```
Killed process 9823, UID 0, (FlowCon.fresher) total-vm:3098244kB, anon-rss:1157280kB, file-rss:36kB
```

### DIFF
--- a/subtree_oom.go
+++ b/subtree_oom.go
@@ -3,6 +3,7 @@ package jasper
 import (
 	"context"
 	"os/exec"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


subtree_oom.go
```diff
@@ -49,25 +50,25 @@ func dmesgContainsOOMKill(line string) bool {
 }
 
 func getPidFromDmesg(line string) (int, bool) {
-	split := strings.Split(line, "Killed process")
-	if len(split) <= 1 {
+	r := regexp.MustCompile(`Killed process (\d+)`)
+	matches := r.FindStringSubmatch(line)
+	if len(matches) != 2 {
```
ablack12, line 55: Are we sure there won't be another number that might also match, such that we should keep len(matches) <= 1 instead of != 2?

ybrill: I was thinking it would always return either a slice of length 2 or nil.

The [docs](https://golang.org/pkg/regexp/#Regexp.FindStringSubmatch) say the slice returned will contain the leftmost match ("Killed process 1234") and the matches of its subexpressions ("1234"). I understood that because there's only one subexpression (`(\d+)`) when there is a match it will return a slice of length 2 and when there's no match it will return nil.

PR #199, opened 2020-07-13
EVG-12502: support generated configs as YAML

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-12502

* Optionally marshal the generated evergreen config as YAML instead of JSON based on CLI flag.
* Include display name (required) for variants.


cli/generate.go
```diff
@@ -24,11 +27,30 @@ func Generate() cli.Command {
...
 }
 
+const (
+	jsonFormat = "json"
+	yamlFormat = "yaml"
+)
+
+type configFormatter func(*shrub.Configuration) ([]byte, error)
+
+func generatedConfigFormats() map[string]configFormatter {
```
julianedwards, line 37: should this be `generateConfigFormatters()` instead?

cli/generate.go
```diff
@@ -168,19 +189,45 @@ func generateMake() cli.Command {
...
-			if err != nil {
-				return errors.Wrap(err, "marshalling evergreen config as JSON")
-			}
-			if outputFile != "" {
-				if err := ioutil.WriteFile(outputFile, output, 0644); err != nil {
-					return errors.Wrapf(err, "writing JSON config to file '%s'", outputFile)
-				}
-			} else {
-				fmt.Println(string(output))
+			if err := formatAndOutputGeneratedConfig(conf, c.String(outputFormatFlagName), outputFile); err != nil {
```
julianedwards, line 192: I think you are missing the `outputFormatFlagName` from the flags in this cli func

cli/generate.go
```diff
@@ -168,19 +189,45 @@ func generateMake() cli.Command {
...
+}
+
+func checkGeneratedConfigFormat(c *cli.Context) error {
+	format := strings.ToLower(c.String(outputFormatFlagName))
+	var formats []string
+	for f := range generatedConfigFormats() {
+		if format == f {
+			return nil
+		}
+		formats = append(formats, format)
```
julianedwards, line 230: This seems unnecessary, can't we just do a lookup on the hashmap for the format we are trying to validate? And if we really want to return the acceptable formats upon error, can't we just have slice somewhere with the list of acceptable formats?

cli/generate.go
```diff
@@ -168,19 +189,45 @@ func generateMake() cli.Command {
...
+				return errors.Wrap(err, "formatting and writing output")
 			}
 
 			return nil
 		},
 	}
 }
+
+func formatAndOutputGeneratedConfig(conf *shrub.Configuration, format, file string) error {
+	doFormatting, ok := generatedConfigFormats()[format]
```
julianedwards, line 202: `strings.ToLower(format)`

cli/generate.go
Kimchelly: I changed this to just take the format as a parameters and switch-case on it.

cli/generate.go
Kimchelly: Fixed.

cli/generate.go
Kimchelly: Moved this to `generatedConfigFormat`.

cli/generate.go
Kimchelly: Fixed.

PR #202, opened 2020-07-15
EVG-12538: Update protobuf file to reflect changes in buildlogger client

julianedwards:


Kimchelly:
You're going to have to restart all the agents in evergreen for this to work since this is another backwards-incompatible change.

PR #205, opened 2020-07-16
EVG-12544 fix hang

ybrill:
My changes from #203 were causing the OOM check to hang. We were indefinitely `select` waiting on cmd.Wait() to write into a channel because the channel wasn't buffered and no one was reading from the channel until the select executed. I made the channel buffered so the write won't block.

This example program convinced me this was the problem:
```
func main() {
	cmd := exec.Command("ls")

	logPipe, err := cmd.StdoutPipe()
	if err != nil {
		log.Fatal(err)
	}

	scanner := bufio.NewScanner(logPipe)
	if err := cmd.Start(); err != nil {
		log.Fatal(err)
	}
	for scanner.Scan() {
		fmt.Println(scanner.Text())
	}

	errs := make(chan error)
	select {
	case errs <- cmd.Wait():
		err = <-errs
		fmt.Println(err)
		return
	}
}
```

I also noticed I forgot to make the fixes from #203 in the macOS version. I refactored the code so they both use the a common function to analyze the log.


oom_tracker_linux_test.go
```diff
@@ -0,0 +1,21 @@
+// +build linux
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetPids(t *testing.T) {
```
Kimchelly: GetPIDs

oom_tracker_linux.go
```diff
@@ -33,46 +40,21 @@ func (o *oomTrackerImpl) Check(ctx context.Context) error {
...
+	return strings.Contains(line, "Out of memory") ||
+		strings.Contains(line, "Killed process") || strings.Contains(line, "OOM killer") ||
+		strings.Contains(line, "OOM-killer")
+}
 
-	if sudo {
-		cmd = exec.CommandContext(ctx, "sudo", "dmesg")
-	} else {
-		cmd = exec.CommandContext(ctx, "dmesg")
+func getPidFromDmesg(line string) (int, bool) {
```
Kimchelly, line 49: getPIDfromDmesg

oom_tracker.go
```diff
@@ -43,34 +41,51 @@ func isSudo(ctx context.Context) (bool, error) {
...
 }
 
-func dmesgContainsOOMKill(line string) bool {
-	return strings.Contains(line, "Out of memory") ||
-		strings.Contains(line, "Killed process") || strings.Contains(line, "OOM killer") ||
-		strings.Contains(line, "OOM-killer")
+type logAnalyzer struct {
+	cmdArgs        []string
+	lineHasOOMKill func(string) bool
+	extractPid     func(string) (int, bool)
```
Kimchelly, line 47: extractPID

oom_tracker_darwin.go
```diff
@@ -35,56 +40,19 @@ func (o *oomTrackerImpl) Check(ctx context.Context) error {
...
 }
 
-func analyzeLogs(ctx context.Context) (bool, []int, error) {
-	var cmd *exec.Cmd
-	wasOOMKilled := false
-	errs := make(chan error)
-	sudo, err := isSudo(ctx)
-	if err != nil {
-		return false, nil, errors.Wrap(err, "error checking sudo")
+func getPidFromLog(line string) (int, bool) {
```
Kimchelly, line 43: getPIDFromLog

oom_tracker_darwin_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+package jasper
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetPidFromLog(t *testing.T) {
```
Kimchelly, line 11: TestGetPIDFromLog

oom_tracker_linux_test.go
ybrill: this has gone away

oom_tracker_linux.go
ybrill: done

oom_tracker.go
ybrill: done

oom_tracker_darwin.go
ybrill: done

oom_tracker_darwin_test.go
ybrill: done

ybrill:
I think I got all of them

PR #208, opened 2020-07-21
EVG-12467: implement SendMessages for CLI

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-12467

* Implement SendMessages for the CLI and CLI over SSH.
* Add validation for correct logging payloads.
* Fix some misnaming in mongo wire protocol logging.
* We can't test this until [this ticket](https://jira.mongodb.org/browse/EVG-12626) is completed since `SendMessages()` relies on the logging cache.

### DIFF
--- a/remote/internal/service.go
+++ b/remote/internal/service.go
@@ -942,19 +942,12 @@ func (s *jasperService) SendMessages(ctx context.Context, lp *LoggingPayload) (*
 	}
 
 	payload := lp.Export()
-	if err := payload.Validate(); err != nil {
-		return &OperationOutcome{
-			Success:  false,
-			Text:     err.Error(),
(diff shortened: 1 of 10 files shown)


remote/internal/service.go
```diff
@@ -942,13 +942,21 @@ func (s *jasperService) SendMessages(ctx context.Context, lp *LoggingPayload) (*
 	}
 
 	payload := lp.Export()
-	if err := logger.Send(payload); err != nil {
+	if err := payload.Validate(); err != nil {
```
julianedwards, line 945: Won't `payload.Send` check validate?

Kimchelly: It will. I removed it.

PR #214, opened 2020-09-08
EVG-12896: use GRPC error codes

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-12896

The code inconsistently uses a mix of just returning internal errors and returning objects containing success/failure to indicate errors. Instead, use gRPC error codes where possible.

This may have some effect on the Jasper resmoke integration since it changes error handling. It seems like curator is pinned to a specific git revision in resmoke, but this might break the existing error-checking code when it gets upgraded.


remote/internal/service.go
```diff
@@ -191,12 +199,12 @@ func (s *jasperService) Group(t *TagName, stream JasperProcessManager_GroupServe
 func (s *jasperService) Get(ctx context.Context, id *JasperProcessID) (*ProcessInfo, error) {
 	proc, err := s.manager.Get(ctx, id.Value)
 	if err != nil {
-		return nil, errors.Wrapf(err, "problem fetching process '%s'", id.Value)
+		return nil, newGRPCError(codes.NotFound, errors.Wrapf(err, "getting process with id '%s'", id.Value))
```
julianedwards, line 202: Couldn't this also be internal? Is it possible to check if it is a not found error?

julianedwards (changes requested):

remote/internal/service.go
Kimchelly: It could theoretically be internal, but the way the base Jasper manager is implemented, the only possible way for it to error is if [the map lookup fails](https://github.com/mongodb/jasper/blob/8c17f722c9d7322a2b929d027bca6cd206295686/manager_basic.go#L167-L170), so I think it's okay to assume that Get will only fail if the process doesn't exist.

julianedwards (approved):

PR #215, opened 2020-09-09
EVG-12896: downgrade packages to ensure they use protobuf v3

Kimchelly:
This is kind of silly, but this avoids a bunch of stupidity in the upgrade to protobuf version 4, like the duplicate vendored dependencies.

I had to fix a bunch of dependency versions to ensure that they're before protobuf upgrades to version 4.


julianedwards (approved):
Will this be trivial to upgrade once we have modules?

Kimchelly:
Yes, modules are intended to handle the transient dependencies without us having to deal with them. The problem is, a lot of projects (e.g. all the gRPC projects) have moved on to modules, which glide can't support.

PR #225, opened 2020-11-30
EVG-13468: Add BSON log format

julianedwards:


Kimchelly (approved):
Can you revendor this in Curator and Evergreen and update the Curator hash in the Evergreen admin settings? This is a backwards-incompatible change.

julianedwards:
Yeah I will do that!

PR #227, opened 2020-12-01
EVG-13492: Change process_name to proc_name for buildloggerv3 options in jasper protobuf

julianedwards:


Kimchelly (commented):
Why do we need to do this? It seems like this will just require us to revendor Jasper in Evergreen again.

julianedwards:
Yeah it is annoying but it is necessary. With the injectable logging backend in jasper, the grpc protobuf needs to match the json or bson tags in the logging backend (timber & cedar in this case). The tag is `proc_name` not `process_name` so that field doesn't get populated in the DB with the current protobuf.

Kimchelly (approved):

PR #229, opened 2020-12-09
EVG-13532: send SIGKILL in Linux process tracker cleanup

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13532

* Only send SIGKILL if SIGTERM fails when cleaning up tracked Linux processes.
* Minor fix for converting int (i.e. char) to string since go1.15 will complain.
* Minor doc comment fixes.

### DIFF
--- a/remote/process_test.go
+++ b/remote/process_test.go
@@ -177,7 +177,7 @@ func addBasicProcessTests(tests ...processTestCase) []processTestCase {
 				opts := testutil.SleepCreateOpts(3)
 				proc, err := makep(ctx, opts)
 				require.NoError(t, err)
-				assert.Error(t, proc.RegisterSignalTriggerID(ctx, jasper.SignalTriggerID(-1)))
+				assert.Error(t, proc.RegisterSignalTriggerID(ctx, jasper.SignalTriggerID("")))
 			},
 		},
 		{
--- a/tracker_linux.go
+++ b/tracker_linux.go
@@ -141,12 +141,13 @@ func (t *linuxProcessTracker) doCleanupByEnvironmentVariable() error {
 // already terminated, this will not return an error.
 func cleanupProcess(pid int) error {
 	// A process returns syscall.ESRCH if it already terminated.
-	catcher := grip.NewBasicCatcher()
-	err := syscall.Kill(pid, syscall.SIGTERM)
-	catcher.AddWhen(err != syscall.ESRCH, errors.Wrapf(err, "sending sigterm to process with PID '%d'", pid))
-	err = syscall.Kill(pid, syscall.SIGKILL)
-	catcher.AddWhen(err != syscall.ESRCH, errors.Wrapf(err, "sending sigkill to process with PID '%d'", pid))
-	return catcher.Resolve()
+	if err := syscall.Kill(pid, syscall.SIGTERM); err != nil && err != syscall.ESRCH {
+		catcher := grip.NewBasicCatcher()
+		catcher.Add(errors.Wrapf(err, "sending sigterm to process with PID '%d'", pid))
+		catcher.Add(errors.Wrapf(syscall.Kill(pid, syscall.SIGKILL), "sending sigkill to process with PID '%d'", pid))
+		return catcher.Resolve()
+	}
+	return nil
 }
 
 // Cleanup kills all tracked processes. If cgroups is available, it kills all
--- a/tracker_linux.go
+++ b/tracker_linux.go
@@ -141,13 +141,12 @@ func (t *linuxProcessTracker) doCleanupByEnvironmentVariable() error {
 // already terminated, this will not return an error.
 func cleanupProcess(pid int) error {
 	// A process returns syscall.ESRCH if it already terminated.
-	if err := syscall.Kill(pid, syscall.SIGTERM); err != nil && err != syscall.ESRCH {
-		catcher := grip.NewBasicCatcher()
-		catcher.Add(errors.Wrapf(err, "sending sigterm to process with PID '%d'", pid))
-		catcher.Add(errors.Wrapf(syscall.Kill(pid, syscall.SIGKILL), "sending sigkill to process with PID '%d'", pid))
-		return catcher.Resolve()
-	}
-	return nil
+	catcher := grip.NewBasicCatcher()
+	err := syscall.Kill(pid, syscall.SIGTERM)
+	catcher.AddWhen(err != syscall.ESRCH, errors.Wrapf(err, "sending sigterm to process with PID '%d'", pid))
+	err = syscall.Kill(pid, syscall.SIGKILL)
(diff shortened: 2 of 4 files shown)


tracker_linux.go
```diff
@@ -140,13 +140,14 @@ func (t *linuxProcessTracker) doCleanupByEnvironmentVariable() error {
...
 func cleanupProcess(pid int) error {
-	catcher := grip.NewBasicCatcher()
 	// A process returns syscall.ESRCH if it already terminated.
-	if err := syscall.Kill(pid, syscall.SIGTERM); err != syscall.ESRCH {
-		catcher.Add(errors.Wrapf(err, "failed to send sigterm to process with pid '%d'", pid))
-		catcher.Add(errors.Wrapf(syscall.Kill(pid, syscall.SIGKILL), "failed to send sigkill to process with pid '%d'", pid))
+	if err := syscall.Kill(pid, syscall.SIGTERM); err != nil && err != syscall.ESRCH {
+		catcher := grip.NewBasicCatcher()
+		catcher.Add(errors.Wrapf(err, "sending sigterm to process with PID '%d'", pid))
+		catcher.Add(errors.Wrapf(syscall.Kill(pid, syscall.SIGKILL), "sending sigkill to process with PID '%d'", pid))
```
ybrill, line 147: should we give the process time to cleanup before we follow up with SIGKILL?

tracker_linux.go
```diff
@@ -140,13 +140,14 @@ func (t *linuxProcessTracker) doCleanupByEnvironmentVariable() error {
 // cleanupProcess terminates the process given by its PID. If the process has
 // already terminated, this will not return an error.
 func cleanupProcess(pid int) error {
-	catcher := grip.NewBasicCatcher()
 	// A process returns syscall.ESRCH if it already terminated.
-	if err := syscall.Kill(pid, syscall.SIGTERM); err != syscall.ESRCH {
-		catcher.Add(errors.Wrapf(err, "failed to send sigterm to process with pid '%d'", pid))
-		catcher.Add(errors.Wrapf(syscall.Kill(pid, syscall.SIGKILL), "failed to send sigkill to process with pid '%d'", pid))
+	if err := syscall.Kill(pid, syscall.SIGTERM); err != nil && err != syscall.ESRCH {
```
ybrill, line 144: if sending a SIGTERM errors why would we be able to send a SIGKILL?

remote/process_test.go
```diff
@@ -177,7 +177,7 @@ func addBasicProcessTests(tests ...processTestCase) []processTestCase {
 				opts := testutil.SleepCreateOpts(3)
 				proc, err := makep(ctx, opts)
 				require.NoError(t, err)
-				assert.Error(t, proc.RegisterSignalTriggerID(ctx, jasper.SignalTriggerID(-1)))
```
ybrill, line 180: pre go1.15 this was Kosher?

tracker_linux.go
Kimchelly: No, I don't think it's worth waiting. Cleanup means kill everything as soon as possible.

tracker_linux.go
Kimchelly: SIGTERM can be ignored, SIGKILL can't.

remote/process_test.go
Kimchelly: Yes. An int is treated as a char, which is then converted to a string.

tracker_linux.go
ybrill: I was assuming syscall.Kill will return an error only if there's a problem _sending the signal_, in which case we wouldn't be able to send SIGKILL either. 
In my mental model syscall.Kill just sends the signal doesn't wait to find out what the process does with it.

tracker_linux.go
ybrill: it there a reason not to SIGKILL as a first resort, then?

Kimchelly: Because SIGKILL won't clean up resources, but SIGTERM can be ignored. https://major.io/2010/03/18/sigterm-vs-sigkill/

tracker_linux.go
Kimchelly: True. It's easier to just send both SIGTERM and SIGKILL and ignore ESRCH in both cases, so I just changed it to ignore ESRCH for SIGKILL.

tracker_linux.go
```diff
@@ -140,12 +140,12 @@ func (t *linuxProcessTracker) doCleanupByEnvironmentVariable() error {
...
-	catcher := grip.NewBasicCatcher()
 	// A process returns syscall.ESRCH if it already terminated.
-	if err := syscall.Kill(pid, syscall.SIGTERM); err != syscall.ESRCH {
-		catcher.Add(errors.Wrapf(err, "failed to send sigterm to process with pid '%d'", pid))
-		catcher.Add(errors.Wrapf(syscall.Kill(pid, syscall.SIGKILL), "failed to send sigkill to process with pid '%d'", pid))
-	}
+	catcher := grip.NewBasicCatcher()
+	err := syscall.Kill(pid, syscall.SIGTERM)
+	catcher.AddWhen(err != syscall.ESRCH, errors.Wrapf(err, "sending sigterm to process with PID '%d'", pid))
+	err = syscall.Kill(pid, syscall.SIGKILL)
```
julianedwards, line 147: Shouldn't we check that there was an error before calling this?

julianedwards (changes requested):

tracker_linux.go
Kimchelly: We could, but we're mostly interested in killing the process. Since the process can ignore SIGTERM, we send SIGKILL in case it doesn't really want to shut down.

tracker_linux.go
ybrill: That's what I was getting at in the beginning, though: if we send the two signals in quick succession the process won't have a chance to clean up before it gets terminated.
As an aside, I found [this comic](https://turnoff.us/geek/dont-sigkill/) simultaneously humorous and terrible.

Kimchelly: Then I guess this begs the question - if we should just trust that processes will obey SIGTERM, we can just send SIGTERM and no error means it's dead. I think that's a reasonable solution, which was the one I originally wrote (i.e. no error from SIGTERM means it's dead).

ybrill: I don't think we can _trust_ processes to obey SIGTERM: the idea is to give them a chance to do cleanup as a curtesy. We can't **know** they'll be dead until we've sent SIGKILL.

Kimchelly: Sure, but this operation is supposed to be fast. If we have to wait for an unknown amount of time for the process to finish, it'll make this operation expensive to use.

julianedwards (approved):

julianedwards:
Can you change the title? It is misleading since we send SIGKILL no matter what.

ybrill (approved):

PR #230, opened 2020-12-14
EVG-13100: add errors for logging cache methods

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13100

* Add errors for logging cache methods.
    * Return an error if `Remove()` is given a logger ID that doesn't exist in the cache.
* Ensure that loggers that are removed via `Clear()` and `Prune()` are closed to avoid resource leaks.
    * If you don't want to close the logger or the logger will not close but you still want to remove it, you can still forcibly remove the logger with `Remove()`, but closing the logger is preferable.
* Rename `logging_cache_size` command to `logging_cache_len` in wire protocol implementation for consistency with the other remote interfaces.
* Move logging cache/scripting interfaces into their own files for remote interfaces.


mock/logging_cache.go
```diff
@@ -58,23 +58,42 @@ func (c *LoggingCache) Put(id string, logger *options.CachedLogger) error {
...
 }
 
-// Get returns an object from the in-memory logging cache.
-func (c *LoggingCache) Get(id string) *options.CachedLogger { return c.Cache[id] }
+// Get returns an object from the in-memory logging cache. It returns an error
+// if it does not exist in the cache.
+func (c *LoggingCache) Get(id string) (*options.CachedLogger, error) {
+	logger, ok := c.Cache[id]
+	if !ok {
+		return nil, errors.Errorf("logger not found")
```
julianedwards, line 66: Why not use `ErrLoggerCacheNotFound`?

julianedwards, line 66: Why not use `ErrLoggerCacheNotFound`?

julianedwards (approved):
Small nit, otherwise LGTM

mock/logging_cache.go
Kimchelly: Oops, fixed.

PR #233, opened 2021-01-07
EVG-13187: try different ports to start the MDB service

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13187

This fixes an issue where wire protocol tests occasionally fail because the service can't listen on the port we choose on Windows. I ran it a bunch of times and it seems to succeed consistently now.

### DIFF
--- a/remote/mdb_util_test.go
+++ b/remote/mdb_util_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"time"
 
 	"github.com/mongodb/grip"
 	"github.com/mongodb/jasper"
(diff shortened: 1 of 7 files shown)


testutil/http.go
```diff
@@ -31,58 +29,35 @@ func PutHTTPClient(client *http.Client) {
...
 			req, err := http.NewRequest(http.MethodGet, url, nil)
 			if err != nil {
-				timer.Reset(timeoutInterval)
+				timer.Reset(backoff)
 				continue
 			}
 			req = req.WithContext(ctx)
 			resp, err := client.Do(req)
 			if err != nil {
-				timer.Reset(timeoutInterval)
```
ybrill, line 56: why remove the Reset?

remote/rest_util_test.go
```diff
@@ -37,38 +36,45 @@ tryStartService:
 				continue tryStartService
 			}
 
+			srvCtx, srvCancel := context.WithCancel(ctx)
 			go func() {
```
ybrill, line 40: would it work to combine the two goroutines? Something like

```
failedToConnect := make(chan struct{})
go func() {
    srvCtx, srvCancel := context.WithCancel(ctx)
    defer srvCancel()
    grip.Warning(app.Run(srvCtx))

    select {
    case <-ctx.Done():
    case <-failedToConnect:
    }
```

remote/mdb_util_test.go
```diff
@@ -12,30 +13,101 @@ import (
...
+			srvCtx, srvCancel := context.WithCancel(ctx)
+			closeService, err := StartMDBService(srvCtx, mngr, addr)
+			if err != nil {
+				srvCancel()
+				return nil, errors.WithStack(err)
+			}
+
+			failedToConnect := make(chan struct{})
+			go func() {
+				defer func() {
```
ybrill, line 41: the same question about combining the goroutines

remote/mdb_util_test.go
```diff
@@ -12,30 +13,101 @@ import (
...
+				case <-failedToConnect:
+				}
+			}()
 
-	client, err := NewMDBClient(ctx, addr, 0)
-	if err != nil {
-		return nil, errors.WithStack(err)
+			client, err = tryConnectToMDBService(ctx, addr)
+			if err != nil {
+				close(failedToConnect)
```
ybrill, line 53: if we can't connect (for some reason) will we loop infinitely?

testutil/http.go
Kimchelly: Oops that was an oversight. I re-added it.

remote/rest_util_test.go
Kimchelly: We can't do this, because `Run()` is a blocking operation, which would prevent the select from stopping the application when the client fails to connect.

remote/mdb_util_test.go
Kimchelly: No because there's a context done check at the top of the infinite for loop (which is the only case in which we would stop trying to initialize the server/client).

remote/mdb_util_test.go
Kimchelly: This already has just one goroutine, so I didn't change anything.

ybrill (approved):
sorry for the delay!

PR #235, opened 2021-01-20
EVG-13684: write tests for executor interface

Kimchelly:
JIRA: https://jira.mongodb.org/browse/EVG-13684

* Write tests for executor interface (which is the layer below the `jasper.Process` implementations).
* Fix some discrepancies in behavior between the standard library exec library and Docker library executors.
* Move around some testutil options into a new package to avoid circular dependencies.

### DIFF
--- a/internal/executor/docker.go
+++ b/internal/executor/docker.go
@@ -43,10 +43,6 @@ type docker struct {
 	signal   syscall.Signal
 }
 
-type DockerOptions struct {
-	Client *client.Client
-}
-
 // NewDocker returns an Executor that creates a process within a Docker
 // container. Callers are expected to clean up resources by calling Close.
 func NewDocker(ctx context.Context, client *client.Client, platform, image string, args []string) Executor {
--- a/internal/executor/docker.go
+++ b/internal/executor/docker.go
@@ -43,6 +43,10 @@ type docker struct {
 	signal   syscall.Signal
 }
 
+type DockerOptions struct {
+	Client *client.Client
+}
+
 // NewDocker returns an Executor that creates a process within a Docker
 // container. Callers are expected to clean up resources by calling Close.
 func NewDocker(ctx context.Context, client *client.Client, platform, image string, args []string) Executor {
@@ -118,6 +122,8 @@ func (e *docker) Start() error {
 	}
 
 	e.setStatus(Running)
+
+	_ = e.getPID()
 
 	return nil
 }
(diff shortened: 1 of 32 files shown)


testutil/docker.go
```diff
@@ -19,7 +19,10 @@ func IsDockerCase(testName string) bool {
 // SkipDockerIfUnsupported skips the test if the platform is not Linux or the
 // SKIP_DOCKER_TESTS environment variable is explicitly set.
 func SkipDockerIfUnsupported(t *testing.T) {
-	skip, _ := strconv.ParseBool(os.Getenv("SKIP_DOCKER_TESTS"))
+	skip, err := strconv.ParseBool(os.Getenv("SKIP_DOCKER_TESTS"))
```
julianedwards, line 22: Should we add a `require.NoError(err)` here?

testutil/docker.go
```diff
@@ -19,7 +19,10 @@ func IsDockerCase(testName string) bool {
...
+	if err == nil && !skip {
```
julianedwards, line 23: Wouldn't this get taken care of on line 26?

internal/executor/docker.go
```diff
@@ -264,28 +266,40 @@ func (e *docker) Wait() error {
 
 	containerID := e.getContainerID()
 
+	handleContextError := func() (handled bool) {
```
julianedwards, line 269: I thought we don't do named return values?

julianedwards (changes requested):

internal/executor/docker.go
Kimchelly: I mean, I do them when the meaning is non-obvious what the return value is. I think the "no named return values" is not a great rule and should only be considered if you're using named returns values as a means to declare output variables. Otherwise, it's a convenient documentation tool (just like naming your input parameters).

testutil/docker.go
Kimchelly: This is an optional environment variable for which the default is "false" (i.e. run the docker tests), so we don't   check the error when it's not set.

testutil/docker.go
Kimchelly: This handles the case in which you set `SKIP_DOCKER_TEST=false` on a non-Linux machine (e.g. you're on MacOS with `SKIP_DOCKER_TESTS=false` for local testing). L26 this handles the case in which `SKIP_DOCKER_TESTS=true` or you're on a non-Linux machine, which by default skips Docker tests.

julianedwards (approved):

PR #237, opened 2021-02-24
EVG-13734 report OOM lines

ybrill:
Include the actual lines from dmesg so the agent can print them in the logs.
Continue to parse out the PIDs though, since they're sent back to the app server and displayed on the task page.

### DIFF
--- a/mock/oom_tracker.go
+++ b/mock/oom_tracker.go
@@ -29,7 +29,7 @@ func (o *OOMTracker) Clear(context.Context) error {
 	return nil
 }
 
-// Report returns the value of the WasOOMKilled and PIDs fields.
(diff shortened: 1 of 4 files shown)


mock/oom_tracker.go
```diff
@@ -30,6 +30,6 @@ func (o *OOMTracker) Clear(context.Context) error {
 }
 
 // Report returns the value of the WasOOMKilled and PIDs fields.
```
Kimchelly, line 32: Change the doc comment to be Lines.

Kimchelly (approved):
LGTM mod nit.

mock/oom_tracker.go
ybrill: done

PR #240, opened 2021-04-19
EVG-14467 Add M1 Mac build variant

kryptonite303:
This is for the new M1 Macs thee BUILD team received. 
I can't create a patch in Evergreen because these hosts can't be added yet (no Jasper for M1 yet)

### DIFF
--- a/evergreen.yaml
+++ b/evergreen.yaml
@@ -228,6 +228,18 @@ buildvariants:
     tasks:
       - name: test_group
 
+  - name: macos-1100-arm64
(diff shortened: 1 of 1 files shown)


bsamek:
@Kimchelly Does this really need a build variant, or is cross compilation sufficient?

Kimchelly:
Assuming you want this to go into the CLI, you should make this change in [Curator](https://github.com/mongodb/curator) instead, since that's where the binary gets built. You can just add another cross-compile variant.

PR #248, opened 2021-10-05
EVG-15479, EVG-15495: remove GO_BIN_PATH and fix linter PATH

Kimchelly:
* Remove `GO_BIN_PATH`. CI tests now depend on `GOROOT` to tell them which Go version to use and which Go binary to use.
* Set the `GOLANGCI_LINT_CACHE`, which is used to control where golangci-lint caches information. By default, golangci-lint will use the user's cache directory (usually in the home directory). This prevents the CI linter from writing outside of the working directory.
* Fix setting the `PATH` variable for the linter.
* Clean up and standardize the makefile.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,58 +1,35 @@
+# start project configuration
 name := jasper
 buildDir := build
 srcFiles := $(shell find . -name "*.go" -not -path "./$(buildDir)/*" -not -name "*_test.go" -not -path "*\#*")
-testFiles := $(shell find . -name "*.go" -not -path "./$(buildDir)/*" -not -path "*\#*")
-testPackages := $(name) cli remote options mock scripting internal-executor
 allPackages := $(testPackages) remote-internal testutil benchmarks util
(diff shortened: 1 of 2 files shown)


makefile
```diff
@@ -73,10 +50,42 @@ $(buildDir)/$(name): cmd/$(name)/$(name).go $(srcFiles)
...
+# start convenience targets for running tests and coverage tasks on a
+# specific package.
+test-%: $(buildDir)/output.%.test
+	
+coverage-%: $(buildDir)/output.%.coverage
+	
+html-coverage-%: $(buildDir)/output.%.coverage.html
+	
+lint-%: $(buildDir)/output.%.lint
+
```
julianedwards, line 84: Can we remove these newlines?

julianedwards (changes requested):

makefile
Kimchelly: They're required so that make recognizes them as targets.

julianedwards (approved):

PR #261, opened 2021-12-02
EVG-15734: upgrade Docker to v20.10.11

Kimchelly:
Jira: https://jira.mongodb.org/browse/EVG-15734

* Bump Docker to the latest stable version. Since Docker isn't using Go modules, it's always risky to upgrade the Docker library without some caution because we can only have one version of Docker across all our repos.
* Add a field for CPU architecture. We already populate the OS, but Docker's `ContainerCreate` API call now requires you to specify the OS/arch, so we have to pass both of them as part of the options.
* Improve the way options are passed to the Docker executor a bit.

### DIFF
--- a/internal/executor/docker.go
+++ b/internal/executor/docker.go
@@ -46,8 +46,8 @@ type docker struct {
 	signal   syscall.Signal
 }
 
-// DockerOptions represent options for a Docker runtime executor.
-// within a Docker container.
+// DockerOptions represent options for a Docker runtime executor within a Docker
+// container.
 type DockerOptions struct {
 	// Client is the Docker client to use. This is required.
 	Client *client.Client
@@ -59,7 +59,7 @@ type DockerOptions struct {
 	// OS is the operating system that Docker is running in. If unspecified,
 	// this defaults to the runtime GOOS.
 	OS string
-	// OS is the CPU architecture of the machine that Docker is running in. If
+	// Arch is the CPU architecture of the machine that Docker is running in. If
 	// unspecified, this defaults to runtime GOARCH.
 	Arch string
 }
--- a/internal/executor/docker.go
+++ b/internal/executor/docker.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"io/ioutil"
+	"runtime"
 	"sync"
 	"syscall"
 	"time"
@@ -16,6 +17,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/mongodb/grip"
 	"github.com/mongodb/grip/message"
+	specs "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 )
 
@@ -25,8 +27,9 @@ type docker struct {
 	stdout   io.Writer
 	stderr   io.Writer
 
-	image    string
-	platform string
+	image string
+	os    string
+	arch  string
 
 	client *client.Client
 	ctx    context.Context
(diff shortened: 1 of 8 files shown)


options/docker.go
```diff
@@ -27,14 +30,17 @@ func (opts *Docker) Validate() error {
 	catcher := grip.NewBasicCatcher()
 	catcher.NewWhen(opts.Port < 0, "port must be positive value")
 	catcher.NewWhen(opts.Image == "", "Docker image must be specified")
-	if opts.Platform == "" {
-		if PlatformSupportsDocker(runtime.GOOS) {
-			opts.Platform = runtime.GOOS
+	if opts.OS == "" {
```
julianedwards, line 33: Nit: we can clean this up by setting a temporary variable with the OS (unless you really want the two separate error messages):
```
os := opts.OS
if os == "" {
        os = runtime.GOOS
}
if !OSSupportsDocker(os) {
        // handle error
}
```

options/docker.go
```diff
@@ -27,14 +30,17 @@ func (opts *Docker) Validate() error {
...
-			catcher.Errorf("failed to set default platform to current runtime platform '%s' because it is unsupported", opts.Platform)
+			catcher.Errorf("cannot set default OS to current runtime OS '%s' because it is unsupported", opts.OS)
 		}
-	} else if !PlatformSupportsDocker(opts.Platform) {
-		catcher.Errorf("unrecognized platform '%s'", opts.Platform)
+	} else if !OSSupportsDocker(opts.OS) {
+		catcher.Errorf("unrecognized OS '%s'", opts.OS)
+	}
+	if opts.Arch == "" {
+		opts.Arch = runtime.GOARCH
```
julianedwards, line 43: Should the arch only be set to the runtime arch only if we set the OS to the runtime OS?

internal/executor/docker_signals.go
```diff
@@ -48,9 +48,9 @@ func dockerToSyscallSignal(dsig string, platform string) (syscall.Signal, error)
 			return sig, nil
 		}
 	default:
-		return syscall.Signal(-1), errors.Errorf("unrecognized platform '%s'", platform)
+		return syscall.Signal(-1), errors.Errorf("unrecognized os '%s'", os)
```
julianedwards, line 51: `os` -> `OS`

internal/executor/docker_signals.go
```diff
@@ -48,9 +48,9 @@ func dockerToSyscallSignal(dsig string, platform string) (syscall.Signal, error)
...
 	}
-	return syscall.Signal(-1), errors.Errorf("unrecognized signal '%s' for platform '%s'", dsig, platform)
+	return syscall.Signal(-1), errors.Errorf("unrecognized signal '%s' for os '%s'", dsig, os)
```
julianedwards, line 53: `os` -> `OS`

internal/executor/docker.go
```diff
@@ -43,22 +46,60 @@ type docker struct {
 	signal   syscall.Signal
 }
 
+// DockerOptions represent options for a Docker runtime executor.
```
julianedwards, line 49: Remove period.

internal/executor/docker.go
```diff
@@ -43,22 +46,60 @@ type docker struct {
...
+	Client *client.Client
+	// Image is the name of the Docker image. This is required.
+	Image string
+	// Command is the arguments to the command to run in the Docker container.
+	// This is required.
+	Command []string
+	// OS is the operating system that Docker is running in. If unspecified,
+	// this defaults to the runtime GOOS.
+	OS string
+	// OS is the CPU architecture of the machine that Docker is running in. If
```
julianedwards, line 62: `OS` -> `Arch`

julianedwards (changes requested):

internal/executor/docker.go
Kimchelly: Fixed.

internal/executor/docker.go
Kimchelly: Fixed.

internal/executor/docker_signals.go
Kimchelly: Fixed.

internal/executor/docker_signals.go
Kimchelly: Fixed.

options/docker.go
Kimchelly: I think the separate error messages are preferable since you can distinguish between the invalid default and invalid explicitly-specified OS cases. This would also have to unconditionally assign the `os` back to `opts.OS` anyways. I don't feel that strongly, but it seems like the current way preserves more explicit error messages, even though the nested ifs are kind of ugly.

options/docker.go
Kimchelly: I thought about this and while it would be kind of strange for a user to set OS xor arch, I think understanding the default values is simpler if they remain independent. I don't feel that strongly about it either way.

julianedwards (approved):

PR #272, opened 2022-02-11
EVG-16236 revendor grip

ybrill:
[EVG-16236](https://jira.mongodb.org/browse/EVG-16236)

Accommodate changes from https://github.com/mongodb/grip/commit/898ba2888109b89e382fe10b1aed1574fe9fe484

### DIFF
--- a/options/logger.go
+++ b/options/logger.go
@@ -1,6 +1,7 @@
 package options
 
 import (
+	"context"
 	"encoding/json"
 	"time"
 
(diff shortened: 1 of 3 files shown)


options/logger.go
```diff
@@ -353,11 +354,14 @@ func NewSafeSender(baseSender send.Sender, opts BaseOptions) (send.Sender, error
 	}
 
 	sender := &SafeSender{}
+	sender.Sender = baseSender
```
Kimchelly, line 357: I would leave this as it originally was implemented with the if-else since the `Sender` value is really based on `opts.Buffer.Buffered`.

Kimchelly (changes requested):

options/logger.go
ybrill: Would you mind to explain more what you mean that the `Sender` value is really based on opts.Buffer.Buffered and why this necessitates the else?

Kimchelly: The `sender.Sender` variable is set in the original code depending on whether `opts.Buffer.Buffered` is set. My preference is to not set a default value if it's immediately followed by logic to overwrite that default and instead use if-else to decide its actual value because it's easier to reason about the value if it's set in the if-else rather than defaulted and later overwritten.

I don't think this is that big a deal, I just find it easier to read.

ybrill: Ah, I thought maybe I was missing something subtle.
done

Kimchelly (approved):

ybrill:
@Kimchelly am I good to merge this? Are the tests just flaky? The failures don't look related as far as I can tell.

Kimchelly:
I've never seen these particular tests be flaky, but I know they don't depend on the buffered sender. I would auto-file some tickets with the build baron service and someone can fix them later.

ybrill:
Okay! I opened [EVG-16326](https://jira.mongodb.org/browse/EVG-16326), [EVG-16327](https://jira.mongodb.org/browse/EVG-16327) and [EVG-16328](https://jira.mongodb.org/browse/EVG-16328)

PR #336, opened 2023-05-23
EVG-19927: Upgrade go and Evergreen distros.

johndaniels:
This upgrades to go 1.20, and makes sure we use newer versions of our Evergreen distros.

### DIFF
--- a/cli/util.go
+++ b/cli/util.go
@@ -10,7 +10,6 @@ import (
 	"math/rand"
 	"net"
 	"os"
-	"path/filepath"
 	"strings"
 	"time"
 
@@ -54,53 +53,6 @@ func requireStringFlag(name string) cli.BeforeFunc {
 	return func(c *cli.Context) error {
 		if c.String(name) == "" {
 			return errors.Errorf("must specify string for flag '%s'", name)
-		}
-		return nil
-	}
-}
-
-func requireOneFlag(names ...string) cli.BeforeFunc { //nolint:unused
-	return func(c *cli.Context) error {
-		var count int
-		for _, name := range names {
-			if c.IsSet(name) {
-				count++
-			}
-		}
-		if count != 1 {
-			return errors.Errorf("must specify exactly one flag from the following: %s", names)
-		}
-		return nil
-	}
-}
-
-// requireRelativePath verifies that the path flag relPathFlagName is set to a
-// path relative to the directory path set for dirFlagName.
-//
-//nolint:unused
-func requireRelativePath(relPathFlagName, pathFlagName string) cli.BeforeFunc {
-	return func(c *cli.Context) error {
-		relPath := util.ConsistentFilepath(c.String(relPathFlagName))
-		path := util.ConsistentFilepath(c.String(pathFlagName))
-		if filepath.IsAbs(relPath) {
-			if strings.HasPrefix(relPath, path) {
-				relPath = strings.TrimPrefix(relPath, path)
-				return errors.Wrapf(c.Set(relPathFlagName, relPath), "setting flag '%s' to relative path", relPathFlagName)
-			}
-			return errors.Errorf("path '%s' must be relative to the path '%s'", relPath, path)
-		}
-		return nil
-	}
-}
-
-// requireStringSliceFlag verifies that the flag name is set to a non-empty
-// string slice.
-//
-//nolint:unused
-func requireStringSliceFlag(name string) cli.BeforeFunc {
-	return func(c *cli.Context) error {
-		if len(c.StringSlice(name)) == 0 {
-			return errors.Errorf("must specify at least one string for flag '%s'", name)
 		}
 		return nil
 	}
--- a/cli/util.go
+++ b/cli/util.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"math"
 	"math/rand"
 	"net"
@@ -60,7 +59,7 @@ func requireStringFlag(name string) cli.BeforeFunc {
 	}
 }
 
-func requireOneFlag(names ...string) cli.BeforeFunc { //nolint: deadcode
+func requireOneFlag(names ...string) cli.BeforeFunc { //nolint:unused
 	return func(c *cli.Context) error {
 		var count int
 		for _, name := range names {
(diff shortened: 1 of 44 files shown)


mock/process.go
```diff
@@ -19,8 +19,8 @@ type Process struct {
 	WaitExitCode                int
 
 	ProcInfo         jasper.ProcessInfo
-	Triggers         jasper.ProcessTriggerSequence
-	SignalTriggers   jasper.SignalTriggerSequence
+	Triggers         jasper.ProcessTriggerSequence `json:"-"`
```
Kimchelly, line 22: I don't have an objection to this, but why do these two fields need JSON omission tags (and not the others)?

internal/executor/executor_test.go
```diff
@@ -47,6 +49,39 @@ func executorTypes() map[string]executorConstructor {
...
+	reader io.Reader
+	writer io.WriteCloser
+	wg     sync.WaitGroup
+}
+
+func (p *pipeReader) start(t *testing.T) {
+	p.wg.Add(1)
+	go func() {
+		result, err := io.ReadAll(p.reader)
+		p.buffer.Write(result)
```
Kimchelly, line 66: It seems like this should log the error (if any) so that it's trackable during test runtime.

internal/executor/executor_test.go
```diff
@@ -47,6 +49,39 @@ func executorTypes() map[string]executorConstructor {
...
+		assert.NoError(t, err)
```
Kimchelly, line 67: nit: Move the assert one line up.

internal/executor/executor_test.go
```diff
@@ -47,6 +49,39 @@ func executorTypes() map[string]executorConstructor {
...
+		p.wg.Done()
+	}()
+}
+
+func (p *pipeReader) closeAndWait() {
+	p.writer.Close()
```
Kimchelly, line 73: Log or return error.

internal/executor/executor_test.go
```diff
@@ -79,12 +114,14 @@ func executorTestCases() []executorTestCase {
...
-				exec.SetStdout(stdout)
+				pipeReader, writer := makePipeReader()
+				pipeReader.start(t)
+				exec.SetStdout(writer)
 				require.NoError(t, exec.Start())
 				require.NoError(t, exec.Wait())
+				pipeReader.closeAndWait()
 				for _, envVar := range env {
-					assert.Contains(t, stdout.String(), envVar)
+					assert.Contains(t, pipeReader.buffer.String(), envVar)
```
Kimchelly, line 124: My instinct is that, despite the logic that Docker shouldn't be writing to this file anymore, the race detector would potentially flag this as a race to read/write the buffer (I've debugged a test that did almost this exact thing, but it was using the built-in `os/exec` library). Does the `buffer` need a rw lock on it?

evergreen.yaml
```diff
@@ -167,67 +167,50 @@ task_groups:
...
-      SKIP_DOCKER_TESTS: true
-    run_on:
-      - archlinux-new-small
-      - archlinux-new-large
-    tasks:
-      - name: test_group
-
   - name: lint-linux
-    display_name: Lint (Arch Linux)
+    display_name: Lint (Ubuntu 22.04)
```
Kimchelly, line 171: Same nit as the other repos, just call this `Lint`.

evergreen.yaml
```diff
@@ -167,67 +167,50 @@ task_groups:
 #           Buildvariants             #
 #######################################
 buildvariants:
-  - name: race-detector
```
Kimchelly, line 170: I would strongly recommend keeping the race detector because Jasper is a heavily race-sensitive repo due to its usage of goroutines.

mock/process.go
johndaniels: Because ProcessTrigger is a function that can't be serialized. I think the actual behavior is that this just silently gets omitted but the static checker flags it to specifically omit. I can add a comment since this is maybe confusing.

internal/executor/executor_test.go
johndaniels: The reason I ended up switching to the pipe reader/writer setup was actually specifically to fix a detected data race. Before there was a Goroutine in runIOStream that would directly write to the buffers read/write interface, and when we read from it it detected a data race. 

In the new setup, pipeReader uses an io.Pipe(), which is safe (and intended) to have a separate goroutine on the reader/writer end, and makes a local goroutine to consume it and put it in the shared buffer. It also has its own sync.WaitGroup to avoid just replicating the data race elsewhere. The test calls closeAndWait which calls Wait() on the WaitGroup, so the goroutine made by pipeReader needs to terminate and call Done() on its waitGroup, avoiding a read/write data race.

The other approach I considered was using the goroutine safe buffer  here: https://gist.github.com/arkan/5924e155dbb4254b64614069ba0afd81 . That is maybe simpler? Would love to hear any thoughts.

johndaniels: I could also refactor the code to somehow make it possible to wait on the wait group made by runIOStream.... not sure if that's better?

evergreen.yaml
johndaniels: It has been incorporated into the standard Ubuntu 22.04 with `RACE_DETECTOR: true`. Interestingly it caught a race that it hadn't before, hence all my changes with pipeReader.

internal/executor/executor_test.go
johndaniels: I updated the code to make the wait group used by runIOStream accessible on the object. We now explicitly wait for that wait group.

internal/executor/executor_test.go
johndaniels: I updated all of these to assert no error. Does that seem reasonable?

mock/process.go
johndaniels: Added a comment.

cli/util.go
```diff
@@ -95,7 +95,8 @@ func requireRelativePath(relPathFlagName, pathFlagName string) cli.BeforeFunc {
 
 // requireStringSliceFlag verifies that the flag name is set to a non-empty
 // string slice.
-//nolint: deadcode
+//
+//nolint:unused
```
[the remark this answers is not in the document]
johndaniels: These are removed. I had assumed they were being preserved for some reason.

evergreen.yaml
johndaniels: This is removed.

johndaniels: Fixed

mock/process.go
johndaniels: Let me look into why this is being json serialized... although it might be overzealous static checking where it _might_ be serialized as opposed to actually being...

internal/executor/executor_test.go
johndaniels: Since that code already exists.... yes let's do that. Thanks for the advice! Will update.

mock/process.go
johndaniels: It was being json serialized to be used as "invalid" input in cli/manager_test.go. I switched to using an arbitrary invalid map instead, so we don't need the weird json tagging.

internal/executor/executor_test.go
johndaniels: Updated

Kimchelly (approved):
LGTM, sorry for all the back and forth, Jasper is finicky!

PR #382, opened 2024-08-29
DEVPROD-1919: remove SSH library support

Kimchelly:
Jasper has functionality to start processes on a remote host using SSH. It has two options, one is to use the `ssh` binary directly and another is to use a Go library for SSHing. We've never had a use for using the Go library implementation, so it can be removed. Will upgrade Evergreen's Jasper dependency after this is merged.

* Delete SSH library implementation of processes.
* Delete all options and tests related to SSH library implementation.

Unrelated to the change, but the integration test started failing seemingly because the MongoDB download feed contents have changed a bit (probably because 4.0 is really old).
* Upgrade `mongod` version used in integration test.
* Disable the `mongod` integration test on non-Linux platforms. The Jasper integration test is really only used as basic proof that Jasper can run a process that's more complicated than say, `echo` or `ls`.  Since `mongod` has a lot of platform-specific dependencies like libcryptd and the integration test doesn't provide that much more test coverage, it's not really worth the trouble of testing on other platforms.


testutil/options/options.go
```diff
@@ -40,24 +40,35 @@ func SleepCreateOpts(num int) *options.Create {
 // ValidMongoDBDownloadOptions returns valid options for downloading a MongoDB
 // archive file.
 func ValidMongoDBDownloadOptions() options.MongoDBDownload {
-	target := runtime.GOOS
-	if target == "darwin" {
-		target = "osx"
+	return options.MongoDBDownload{
+		BuildOpts: ValidMongoDBBuildOptions(),
+		Releases:  []string{"7.0-current"},
```
malikchaya2, line 45: should we maybe avoid hardcoding `"7.0-current"`?

malikchaya2 (approved):

testutil/options/options.go
Kimchelly: I guess we could move it to a project var, but I'm not sure that's a lot better since it would invisibly make the test run with a different MongoDB version.

So far, this is the only time I've ever had to do this in the many years since it was written, but we can reconsider or possibly remove/rework this test to not need `mongod` if it happens more.

PR #384, opened 2025-02-14
DEVPROD-14445 Add nil check to create validate

bynn:
add nil check for nil stat and check errors

I feel like it looks a bit ugly but wasn't sure if there was a better way

### DIFF
--- a/options/create.go
+++ b/options/create.go
@@ -114,9 +114,14 @@ func (opts *Create) Validate() error {
 
 	if opts.WorkingDirectory != "" && opts.isLocal() {
 		info, err := os.Stat(opts.WorkingDirectory)
-
-		if os.IsNotExist(err) {
-			catcher.Errorf("cannot use '%s' as working directory because it does not exist", opts.WorkingDirectory)
+		if err != nil {
+			if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


options/create.go
```diff
@@ -114,9 +114,14 @@ func (opts *Create) Validate() error {
 
 	if opts.WorkingDirectory != "" && opts.isLocal() {
 		info, err := os.Stat(opts.WorkingDirectory)
-
-		if os.IsNotExist(err) {
-			catcher.Errorf("cannot use '%s' as working directory because it does not exist", opts.WorkingDirectory)
+		if err != nil {
```
Kimchelly, line 117: Since Go evaluates if-else if blocks in top down order and stop at the first true boolean, this can be simplified down to a series of if-else if blocks:
```go
if os.IsNotExist(err) {
    catcher.Errorf(...)
} else if err != nil {
    catcher.Wrapf(err, ...)
} else if info == nil {
    catcher.Errorf(...)
} else if !info.isDir() { // At this point, we already confirmed info is not nil
    catcher.Errorf(...)
}
```

bynn: oh... updated

Kimchelly (approved):

